PR #1, opened 2018-11-08
Drop support for ansible target/initiator/LUN configuration

dillaman:
Configuration will be handled by 'gwcli' and the Ceph Dashboard.

Signed-off-by: Jason Dillaman <dillaman@redhat.com>


dillaman:
@mikechristie Yeah, I think going forward it would be best to remove the ```manage``` functions.

PR #3, opened 2018-11-27
Fix errors when performing operations on invalid gateway

ricardoasmarques:
This PR will fix the following error when performing a disk operation on a machine that is not a valid gateway:

`Failed : Unhandled exception: list.remove(x): x not in list`

**How to reproduce**

Given the following configuration:

```
  o- disks .............................................................. [0.00Y, Disks: 0]
  o- iscsi-target ............................................................ [Targets: 1]
    o- iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw ............................ [Gateways: 2]
      o- gateways ................................................... [Up: 2/2, Portals: 2]
      | o- node2 ................................................... [192.168.100.202 (UP)]
      | o- node3 ................................................... [192.168.100.203 (UP)]
      o- host-groups ......................................................... [Groups : 0]
      o- hosts ..................................................... [Hosts: 0: Auth: None]
```
Execute the following on `node1`:
```
[@node1] gwcli
/> cd disks 
/disks> create pool=rbd image=disk_1 size=90G
Failed : Unhandled exception: list.remove(x): x not in list
```

### DIFF
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -35,8 +35,9 @@ from ceph_iscsi_config.utils import (normalize_ip_literal, resolve_ip_addresses,
                                      ip_addresses, gen_file_hash, valid_rpm,
                                      CephiSCSIError)
 
-from gwcli.utils import (this_host, APIRequest, valid_gateway,
-                         valid_client, valid_snapshot_name, GatewayAPIError)
+from gwcli.utils import (this_host, APIRequest, valid_gateway, valid_client,
+                         get_remote_gateways, valid_snapshot_name,
+                         GatewayAPIError)
 
 app = Flask(__name__)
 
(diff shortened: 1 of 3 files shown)


rbd-target-api.py
```diff
@@ -688,6 +688,15 @@ def disk(image_id):
...
         return jsonify(message=msg), 400
 
+    # Any disk operation must be started by the `local_gw` (owner)
+    # so `local_gw` must exist in the gateways configuration
+    if local_gw not in gateways:
+        msg = "local machine ({}) cannot be used to perform disk " \
+              "operations because it is not defined within gateways " \
+              "configuration".format(local_gw)
+        logger.warning("disk create request failed: {}".format(msg))
+        return jsonify(message=msg), 400
```
mikechristie: It looks like this is a requirement for multiple operations. Check out the target, client, etc operations.

Move this to a new function so we can call in the other operations that need the same check.

ricardoasmarques: @mikechristie I've applied the validation to all operations where it's needed

mikechristie:
I think we can merge a little more common code since we really want to start with just the remove gws. How about the attached patch?

[ceph-iscsi-get-gw-list.txt](https://github.com/ceph/ceph-iscsi/files/2631350/ceph-iscsi-get-gw-list.txt)

mikechristie:
Here is the correct patch.

[ceph-iscsi-get-gw-list-v2.txt](https://github.com/ceph/ceph-iscsi/files/2631359/ceph-iscsi-get-gw-list-v2.txt)

ricardoasmarques:
Thanks @mikechristie, I've applied your patch v2 with the following additional change:

```diff
diff --git a/rbd-target-api.py b/rbd-target-api.py
index 5f22569..dc09a6c 100755
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -684,7 +684,7 @@ def disk(image_id):
         return jsonify(message="{}".format(err)), 400
 
     # Any disk operation needs at least 2 gateways to be present
-    if len(gateways) < settings.config.minimum_gateways:
+    if len(gateways) + 1 < settings.config.minimum_gateways:
         msg = "at least {} gateways must exist before disk operations " \
               "are permitted".format(settings.config.minimum_gateways)
         logger.warning("disk create request failed: {}".format(msg))
```

PR #5, opened 2018-12-04
Fix ALUATargetPortGroup property name

ricardoasmarques:
`ALUATargetPortGroup` doesn't have a property named `tpg_id`, the correct property to get the ALUA target port group id is `tg_pt_gp_id`.

### DIFF
--- a/ceph_iscsi_config/gateway.py
+++ b/ceph_iscsi_config/gateway.py
@@ -375,7 +375,7 @@ class GWTarget(GWObject):
             # someone mapped a LU then unmapped it without deleting the
             # stg_object, or we are reloading the config.
             alua_tpg = ALUATargetPortGroup(stg_object, group_name)
-            if alua_tpg.tpg_id != tpg.tag:
+            if alua_tpg.tg_pt_gp_id != tpg.tag:
                 # ports and owner were rearranged. Not sure we support that.
                 raise CephiSCSIInval("Existing ALUA group tag for group {} "
                                      "in invalid state.\n".format(group_name))


ceph_iscsi_config/gateway.py
```diff
@@ -375,7 +375,7 @@ def bind_alua_group_to_lun(self, config, lun, tpg_ip_address=None):
             # someone mapped a LU then unmapped it without deleting the
             # stg_object, or we are reloading the config.
             alua_tpg = ALUATargetPortGroup(stg_object, group_name)
-            if alua_tpg.tpg_id != tpg.tag:
+            if alua_tpg.tg_pt_gp_id != tpg.tag:
```
mikechristie: I just fixed that and the bug that is hit when we go through that path here:

https://github.com/mikechristie/ceph-iscsi-config/commit/f360a4d12837e0fc3a6ffcc9705e7bf9b8cfaaf4

I am still testing it so I had not done a pull yet.

mikechristie: Are you hitting the error described in my commit or did you find this just reviewing the code?

mikechristie: Sorry for all the emails.

I am giving myself a review comment for my patch to catch errors and propagate them up properly instead of crashing.

ricardoasmarques: I found it because this error is thrown on my current multi-target implementation when the same host is used on different targets (I have to check if this is supposed to happen or it's a bug on my implementation):

screenshot from 2018-12-06 01-26-25

I've tested your solution and apparently it fixes the issue.

mikechristie: Let me fix up my patch. I think it might be wrong for what you are going to want to do with multiple target support.

My patch was handling the case where rbd-target-gw crashes, then on restart it is detecting an existing and possibly stale config.

For multiple targets if you were to map a disk/image to N targets then it is expected that the alua group will already be setup in the kernel after the initial target is setup, and we do not want to resetup it up over and over.

ricardoasmarques: @mikechristie just to clarify, is there anything I can do in this PR in order to unblock https://github.com/ceph/ceph-iscsi/pull/9, or should I just wait for your patch?

mikechristie: Sorry. I have been debating with myself. I am thinking that maybe when rbd-target-gw starts if it finds an existing setup maybe it should clear it, because it does not know what has changed and the entire system could be out of sync.

In that case we just want your patch.

Thoughts?

ricardoasmarques:
I'm closing this PR because this commit is already included in https://github.com/ceph/ceph-iscsi/pull/9

PR #7, opened 2018-12-06
Centralize the code that upgrades the config file

ricardoasmarques:
In preparation to a more complex upgrade from config v3 to v4, this PR will centralize the existing upgrade code into a single place where we will then implement future config upgrades. 


ricardoasmarques:
I'm closing this PR because this commit is already included in https://github.com/ceph/ceph-iscsi/pull/9

Issue #6, opened 2018-12-06
Migrate away from pycrypto?


smithfarm:
Hi. I noticed that ceph-iscsi has a runtime dependency on pycrypto 2.6.1 [1]

This is problematic for openSUSE because this package has been removed from the distribution. It is slated to be dropped from SLES as well, though that has not happened just yet.

Apparently, this package is only used here: https://github.com/ceph/ceph-iscsi/blob/c6860ae477dff72239b5f3df45ae5fe48e17d308/ceph_iscsi_config/client.py#L3

The openSUSE maintainers suggest that we migrate to https://pypi.org/project/pycryptodomex/ or https://pypi.org/project/pyOpenSSL/ - is this feasible?

[1] https://www.dlitz.net/software/pycrypto/

dillaman:
I'd vote for pyOpenSSL since Ceph already has a dependency on it.

rjfd:
@dillaman @smithfarm the pyOpenSSL only works for managing keys/certificates it does not have the encrypt/decrypt function that it's required by the ceph-iscsi code.

In the the pyOpenSSL documentation they recommend the usage of python-cryptography package to preform encrypt/decrypt (https://pyopenssl.org/en/stable/api/crypto.html#module-OpenSSL.crypto).

Can we use python-cryptography instead?

smithfarm:
@rjfd Works for me:

SLE-15: python3-cryptography 2.1.4
openSUSE:Factory: python3-cryptography 2.3.1

dillaman:
That's fine w/ me as well since it's available in RHEL/CentOS

PR #8, opened 2018-12-06
chap: migrate from python-crypt to python-cryptography

rjfd:

### DIFF
--- a/ceph_iscsi_config/client.py
+++ b/ceph_iscsi_config/client.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 
-from Crypto.PublicKey import RSA
-from Crypto.Cipher import PKCS1_OAEP
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization, hashes
+from cryptography.hazmat.primitives.asymmetric import padding
 from base64 import b64encode, b64decode
 import os
 
@@ -667,7 +668,7 @@ class CHAP(object):
 
     def _set_chap_str(self, chap_str):
 
-        username, self.password_str = chap_str.split('/', 1)
+        self.user, self.password_str = chap_str.split('/', 1)
 
         if encryption_available() and len(self.password_str) > 0:
             self.password = self._encrypt()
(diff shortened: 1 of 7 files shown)


ceph_iscsi_config/client.py
```diff
@@ -679,10 +680,15 @@ def _decrypt(self):
...
         try:
-            key = RSA.importKey(open(key_path))
-
-            cipher = PKCS1_OAEP.new(key)
-            plain_pw = cipher.decrypt(b64decode(self.password_str))
+            with open(key_path, 'rb') as keyf:
+                key = serialization.load_pem_private_key(keyf.read(), None,
+                                                         default_backend())
+
+            plain_pw = key.decrypt(b64decode(self.password_str),
```
dillaman: Nit: I might be wrong, but I think the older library might have defaulted to the insecure SHA-1 hash. If the CHAP password was previously encrypted w/ the SHA-1 hash, would it fail to decode here? If so, for initial upgrade backwards compatibility, should it catch ```ValueError``` and try again w/ the downgrade?

rjfd: I think you might be right. I forgot about the upgrade path. I'll add a unit test that encrypts the password using the previous library (python-crypto) and decrypt with the new code. That will give us some confidence.

rjfd:
@dillaman the code can now handle the case where the password had been encrypted with the previous code

rjfd:
looks like jenkins slave does not have python3-devel package installed, and pip is unable to build pycrypto for python 3 environment.

LenzGr:
FYI: this Fixes github issue #6 (would it make sense that ceph-iscsi issues are tracked on the Ceph bug tracker as well?)

dillaman:
@rjfd Perhaps just re-use python-cryptography but encrypt it in your test using:

```
            encrypted_pw = b64encode(key.encrypt(self.password_str.encode('utf-8'),
                                     padding.OAEP(
                                         mgf=padding.MGF1(algorithm=hashes.SHA1()),
                                         algorithm=hashes.SHA1(),
                                         label=None))).decode('utf-8')

```

rjfd:
@dillaman done

PR #9, opened 2018-12-10
Multiple iSCSI target support (and more...)

ricardoasmarques:
This PR adds support for multiple iSCSI targets.

screenshot from 2018-12-10 14-48-50

**Some other changes included in this PR:**

- Add `detach <disk>` command
      (similar to `delete <disk>` but will not delete the RBD image)

- Add `attach <disk>` command
      (similar to `create <disk>` but will not create the RBD image)

- Lazy creation of LIO storage objects
      (LIO storage objects will only be created when mapped to a target)

- Allow disk deletion from a `gwcli` running on any host
      (instead of forcing the user to connect to the disk allocating host)

- Mutual chap authentication

- Discovery authentication (chap and chap_mutual)

### DIFF
--- a/ceph_iscsi_config/client.py
+++ b/ceph_iscsi_config/client.py
@@ -29,18 +29,19 @@ class GWClient(GWObject):
                 "nopin_timeout",
                 "cmdsn_depth"]
 
-    seed_metadata = {"auth": {"chap": ''},
+    seed_metadata = {"auth": {"chap": '', "chap_mutual": ''},
                      "luns": {},
                      "group_name": ""
                      }
 
-    def __init__(self, logger, client_iqn, image_list, chap, target_iqn):
+    def __init__(self, logger, client_iqn, image_list, chap, chap_mutual, target_iqn):
         """
         Instantiate an instance of an LIO client
         :param client_iqn: (str) iscsi iqn string
         :param image_list: (list) list of rbd images (pool/image) to attach
                            to this client or list of tuples (disk, lunid)
         :param chap: (str) chap credentials in the format 'user/password'
+        :param chap_mutual: (str) chap_mutual credentials in the format 'user/password'
         :param target_iqn: (str) target iqn string
         :return:
         """
@@ -66,6 +67,7 @@ class GWClient(GWObject):
                 self.requested_images = image_list
 
         self.chap = chap                        # parameters for auth
+        self.chap_mutual = chap_mutual          # parameters for mutual auth
         self.mutual = ''
         self.tpgauth = ''
         self.metadata = {}
@@ -235,6 +237,7 @@ class GWClient(GWObject):
         for client_iqn in target_config['clients']:
             client_metadata = target_config['clients'][client_iqn]
             client_chap = CHAP(client_metadata['auth']['chap'])
+            client_chap_mutual = CHAP(client_metadata['auth']['chap_mutual'])
 
             image_list = client_metadata['luns'].keys()
 
@@ -243,11 +246,17 @@ class GWClient(GWObject):
                 raise CephiSCSIError("Unable to decode password for {}. "
                                      "CHAP error: {}".format(client_iqn,
                                                              client_chap.error_msg))
+            chap_mutual_str = client_chap_mutual.chap_str
+            if client_chap_mutual.error:
+                raise CephiSCSIError("Unable to decode password for {}. "
+                                     "CHAP_MUTUAL error: {}".format(client_iqn,
+                                                                    client_chap.error_msg))
 
             client = GWClient(logger,
                               client_iqn,
                               image_list,
                               chap_str,
+                              chap_mutual_str,
                               target_iqn)
 
             client.manage('present')  # ensure the client exists
@@ -266,71 +275,91 @@ class GWClient(GWObject):
 
         tpg.set_attribute('authentication', '0')
 
-    def configure_auth(self, auth_type, credentials):
+    def configure_auth(self, chap_credentials, chap_mutual_credentials):
         """
         Attempt to configure authentication for the client
         :return:
         """
 
-        chap_enabled = False
-        if '/' in credentials:
-            client_username, client_password = credentials.split('/', 1)
-            chap_enabled = True
-        elif not credentials:
+        auth_enabled = False
+        if '/' in chap_credentials:
+            client_username, client_password = chap_credentials.split('/', 1)
+            auth_enabled = True
+        elif not chap_credentials:
             client_username = ''
             client_password = ''
-            credentials = "/"
+            chap_credentials = "/"
 
-        self.logger.debug("configuring auth {}".format(credentials))
+        if '/' in chap_mutual_credentials:
+            client_mutual_username, client_mutual_password = chap_mutual_credentials.split('/', 1)
+            auth_enabled = True
+        elif not chap_mutual_credentials:
+            client_mutual_username = ''
+            client_mutual_password = ''
+            chap_mutual_credentials = "/"
 
-        if auth_type == 'chap':
+        self.logger.debug("configuring auth chap={}, chap_mutual={}".format(
+            chap_credentials, chap_mutual_credentials))
 
-            acl_credentials = "{}/{}".format(self.acl.chap_userid,
-                                             self.acl.chap_password)
+        acl_credentials = "{}/{}".format(self.acl.chap_userid,
+                                         self.acl.chap_password)
+        acl_mutual_credentials = "{}/{}".format(self.acl.chap_mutual_userid,
+                                                self.acl.chap_mutual_password)
 
-            # if the credentials match... nothing to do
-            if credentials == acl_credentials:
-                return
-            else:
-                # credentials defined on the ACL don't match parms passed from
-                # caller so update the acl definition
-                try:
-                    if auth_type == 'chap':
-                        self.logger.debug("Updating the ACL")
-                        self.acl.chap_userid = client_username
-                        self.acl.chap_password = client_password
+        try:
+            self.logger.debug("Updating the ACL")
+            if chap_credentials != acl_credentials:
+                self.acl.chap_userid = client_username
+                self.acl.chap_password = client_password
 
-                        new_chap = CHAP(credentials)
-                        self.logger.debug("chap object set to: {},{},{},{}".format(
-                            new_chap.chap_str, new_chap.user, new_chap.password,
-                            new_chap.password_str))
+                new_chap = CHAP(chap_credentials)
+                self.logger.debug("chap object set to: {},{},{},{}".format(
+                    new_chap.chap_str, new_chap.user, new_chap.password,
+                    new_chap.password_str))
 
-                        new_chap.chap_str = "{}/{}".format(client_username,
-                                                           client_password)
-                        if new_chap.error:
-                            self.error = True
-                            self.error_msg = new_chap.error_msg
-                            return
-
-                        if chap_enabled:
-                            self.tpg.set_attribute('authentication', '1')
-                        else:
-                            self.try_disable_auth(self.tpg)
-
-                        self.logger.debug("Updating config object meta data")
-                        self.metadata['auth']['chap'] = new_chap.chap_str
-
-                except RTSLibError as err:
+                new_chap.chap_str = "{}/{}".format(client_username,
+                                                   client_password)
+                if new_chap.error:
                     self.error = True
-                    self.error_msg = ("Unable to configure {} authentication "
-                                      "for {} - ".format(auth_type,
-                                                         self.iqn,
-                                                         err))
-                    self.logger.error("Client.configure_auth) failed to set "
-                                      "{} credentials for {}".format(auth_type,
-                                                                     self.iqn))
-                else:
-                    self.change_count += 1
+                    self.error_msg = new_chap.error_msg
+                    return
+
+            if chap_mutual_credentials != acl_mutual_credentials:
+                self.acl.chap_mutual_userid = client_mutual_username
+                self.acl.chap_mutual_password = client_mutual_password
+
+                new_chap_mutual = CHAP(chap_mutual_credentials)
+                self.logger.debug("chap mutual object set to: {},{},{},{}".format(
+                    new_chap_mutual.chap_str, new_chap_mutual.user, new_chap_mutual.password,
+                    new_chap_mutual.password_str))
+
+                new_chap_mutual.chap_str = "{}/{}".format(client_mutual_username,
+                                                          client_mutual_password)
+                if new_chap_mutual.error:
+                    self.error = True
+                    self.error_msg = new_chap_mutual.error_msg
+                    return
+
+            if auth_enabled:
+                self.tpg.set_attribute('authentication', '1')
+            else:
+                self.try_disable_auth(self.tpg)
+
+            self.logger.debug("Updating config object meta data")
+            if chap_credentials != acl_credentials:
+                self.metadata['auth']['chap'] = new_chap.chap_str
+            if chap_mutual_credentials != acl_mutual_credentials:
+                self.metadata['auth']['chap_mutual'] = new_chap_mutual.chap_str
+
+        except RTSLibError as err:
+            self.error = True
+            self.error_msg = ("Unable to configure authentication "
+                              "for {} - ".format(self.iqn,
+                                                 err))
+            self.logger.error("(Client.configure_auth) failed to set "
+                              "credentials for {}".format(self.iqn))
+        else:
+            self.change_count += 1
 
     def _add_lun(self, image, lun):
         """
@@ -492,8 +521,16 @@ class GWClient(GWObject):
                     self.error = True
                     self.error_msg = config_chap.error_msg
                     return
+                # extract the chap_mutual_str from the config object entry
+                config_chap_mutual = CHAP(self.metadata['auth']['chap_mutual'])
+                chap_mutual_str = config_chap_mutual.chap_str
+                if config_chap_mutual.error:
+                    self.error = True
+                    self.error_msg = config_chap_mutual.error_msg
+                    return
 
                 if self.chap == chap_str and \
+                   self.chap_mutual == chap_mutual_str and \
                    config_image_list == sorted(self.requested_images):
                     self.commit_enabled = False
             else:
@@ -531,12 +568,11 @@ class GWClient(GWObject):
                                       "for {}".format(bad_images, self.iqn))
                     return
 
-            # if '/' in self.chap:
-            if self.chap == '':
+            if self.chap == '' and self.chap_mutual == '':
                 self.logger.warning("(main) client '{}' configured without"
                                     " security".format(self.iqn))
 
-            self.configure_auth('chap', self.chap)
+            self.configure_auth(self.chap, self.chap_mutual)
             if self.error:
                 return
 
--- a/ceph_iscsi_config/client.py
+++ b/ceph_iscsi_config/client.py
@@ -9,7 +9,8 @@ import os
 import rtslib_fb.root as lio_root
 
 from socket import gethostname
-from rtslib_fb.target import NodeACL, TPG
+from rtslib_fb.target import NodeACL, Target, TPG
+from rtslib_fb.fabric import ISCSIFabricModule
 from rtslib_fb.utils import RTSLibError, normalize_wwn
 
 import ceph_iscsi_config.settings as settings
@@ -178,30 +179,31 @@ class GWClient(GWObject):
         :return:
         """
 
-        r = lio_root.RTSRoot()
+        iscsi_fabric = ISCSIFabricModule()
+        target = Target(iscsi_fabric, self.target_iqn, 'lookup')
 
         # NB. this will check all tpg's for a matching iqn
-        for client in r.node_acls:
-            target = client.parent_tpg.parent_target
-            if client.node_wwn == self.iqn and target.wwn == self.target_iqn:
-                self.acl = client
-                self.tpg = client.parent_tpg
-                try:
-                    self.update_acl_controls()
-                except RTSLibError as err:
-                    self.logger.error("(Client.define_client) FAILED to update "
-                                      "{}".format(self.iqn))
-                    self.error = True
-                    self.error_msg = err
-                self.logger.debug("(Client.define_client) - {} already "
-                                  "defined".format(self.iqn))
-                return
+        for tpg in target.tpgs:
+            for client in tpg.node_acls:
+                if client.node_wwn == self.iqn:
+                    self.acl = client
+                    self.tpg = client.parent_tpg
+                    try:
+                        self.update_acl_controls()
+                    except RTSLibError as err:
+                        self.logger.error("(Client.define_client) FAILED to update "
+                                          "{}".format(self.iqn))
+                        self.error = True
+                        self.error_msg = err
+                    self.logger.debug("(Client.define_client) - {} already "
+                                      "defined".format(self.iqn))
+                    return
 
         # at this point the client does not exist, so create it
         # The configuration only has one active tpg, so pick that one for any
         # acl definitions
-        for tpg in r.tpgs:
-            if tpg.enable and tpg.parent_target.wwn == self.target_iqn:
+        for tpg in target.tpgs:
+            if tpg.enable:
                 self.tpg = tpg
 
         try:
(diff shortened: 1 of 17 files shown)


rbd-target-api.py
```diff
@@ -839,24 +1010,6 @@ def _disk(image_id):
                 return jsonify(message="LUN allocation failure"), 500
 
             if mode == 'create':
-                # new disk is allocated, so refresh the local config object
-                config.refresh()
```
dillaman: This missing ```refresh``` seems to create a race condition w/  creating disks via ```gwcli```:

```
/disks> create pool=rbd image=lun1 size=1G
CMD: /disks/ create pool=rbd image=lun1 size=1G count=1 controls={}
pool 'rbd' is ok to use
Creating/mapping disk rbd/lun1
Issuing disk create request
- LUN(s) ready on all gateways
ok
Updating UI for the new disk(s)
  File "/bin/gwcli", line 194, in <module>
    main()
  File "/bin/gwcli", line 125, in main
    shell.run_interactive()
  File "/usr/lib/python2.7/site-packages/configshell_fb/shell.py", line 905, in run_interactive
    self._cli_loop()
  File "/usr/lib/python2.7/site-packages/configshell_fb/shell.py", line 734, in _cli_loop
    self.run_cmdline(cmdline)
  File "/usr/lib/python2.7/site-packages/configshell_fb/shell.py", line 848, in run_cmdline
    self._execute_command(path, command, pparams, kparams)
  File "/usr/lib/python2.7/site-packages/configshell_fb/shell.py", line 823, in _execute_command
    result = target.execute_command(command, pparams, kparams)
  File "/usr/lib/python2.7/site-packages/configshell_fb/node.py", line 1406, in execute_command
    return method(*pparams, **kparams)
  File "/usr/lib/python2.7/site-packages/gwcli/storage.py", line 245, in ui_command_create
    controls=attributes)
  File "/usr/lib/python2.7/site-packages/gwcli/storage.py", line 337, in create_disk
    "for '{}' from the API".format(disk_key))
gwcli.utils.GatewayAPIError: Unable to retrieve disk details for 'rbd.lun1' from the API
```

ceph_iscsi_config/lun.py
```diff
@@ -241,7 +241,6 @@ def __init__(self, logger, pool, image, size, allocating_host):
         # only uses shortname so it needs to be converted to shortname format
         self.allocating_host = allocating_host.split('.')[0]
 
-        self.owner = ''     # gateway that owns the preferred path for this LUN
```
dillaman: In ```gwcli```, running "/disks info <disk_id>" on an unassigned disk will result in a failure: ```AttributeError: 'Disk' object has no attribute 'owner'```

See https://github.com/ceph/ceph-iscsi/blob/master/gwcli/storage.py#L586

ceph_iscsi_config/lun.py
```diff
@@ -340,6 +368,11 @@ def remove_lun(self):
                                         disk_owner,
                                         gw_metadata)
 
+            disk_metadata = self.config.config['disks'][self.config_key]
+            del disk_metadata['owner']
```
dillaman: Nit: same comment here re: availability of ```owner``` key in CLI

rbd-target-api.py
```diff
@@ -613,6 +632,165 @@ def _gateway(gateway_name=None):
...
+            'disk': disk,
+            'purge_host': local_gw
+        }
+        gateways = target_config.get('ip_list', [])
+        resp_text, resp_code = call_api(gateways, '_targetlun',
+                                        '{}'.format(target_iqn),
+                                        http_method='delete',
+                                        api_vars=api_vars)
+        if resp_code != 200:
+            return jsonify(message="Delte target LUN mapping failed - "
```
dillaman: Nit: typo on ```Delete```

rbd-target-api.py
```diff
@@ -613,6 +632,165 @@ def _gateway(gateway_name=None):
...
+
+        lun = LUN(logger,
+                  pool,
+                  image,
+                  0,
+                  purge_host)
+
+        if lun.error:
+            # problem defining the LUN instance
+            logger.error("Error initialising the LUN : "
```
dillaman: Nit: ```initializing```

rbd-target-api.py
```diff
@@ -876,6 +1029,11 @@ def _disk(image_id):
                 logger.error("LUN size unknown - {}".format(image_id))
                 return jsonify(message="LUN {} failure".format(mode)), 500
 
+            if 'owner' not in disk:
```
dillaman: Nit: when I attempted "/disks reconfigure <disk id> max_data_area_mb 128" against an image not yet assigned to a target, I hit: ```Failed to reconfigure : failed to deactivate disk: failed on ceph-iscsi0. LUN deactivate failure - Disk rbd.lun2 must be added to a target```

We might want a nicer error message like "cannot reconfigure until disk assigned to target". Also, now when creating disks, it probably doesn't make sense to allow any control overrides since they cannot be validated until the disk is mapped to a target in a later step.

ricardoasmarques:
@dillaman your comments were addressed

dillaman:
@ricardoasmarques One quick thing that I forgot, looking at the ```gwcli``` picture above, it shows ```iscsi-target``` with two targets under it -- should we change this to ```iscsi-targets```?

ricardoasmarques:
@dillaman Thanks for the feedback. I've implemented your suggestions.

ricardoasmarques:
rebased

ricardoasmarques:
I've pushed a new commit that adds the support for **single target deletion**.

Additionally, I've pushed some other minor fixes:
- Delete a disk from a target will now validate if the disk is in use by a group
- Prevent error when creating a target that already exists (from the cli)

ceph_iscsi_config/client.py
```diff
@@ -98,7 +100,7 @@ def __init__(self, logger, client_iqn, image_list, chap):
 
         try:
             # We only support global client controls.
```
mikechristie: Drop comment since this is now per target.

ceph_iscsi_config/gateway.py
```diff
@@ -194,15 +194,16 @@ def enable_active_tpg(self, config):
...
         """
         Remove the target definition form LIO
         :return: None
         """
         # check that there aren't any disks or clients in the configuration
         lio_root = root.RTSRoot()
 
-        disk_count = len([disk for disk in lio_root.storage_objects])
+        disk_count = len([disk for disk in lio_root.storage_objects
+                          if disk.name in target_disks or disk.name in unmapped_disks])
```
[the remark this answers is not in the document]
ricardoasmarques: At the time it seemed like a good idea to avoid orphan discs, but after thinking again, it makes sense to remove this restriction. I will fix it.

ceph_iscsi_config/gateway.py
```diff
@@ -96,7 +96,7 @@ def __init__(self, logger, iqn, gateway_ip_list, enable_portal=True):
 
         try:
             # We only support global target/tpg controls.
```
mikechristie: Drop comment since we do per target settings now.

rbd-target-api.py
```diff
@@ -246,88 +246,103 @@ def target(target_iqn=None):
...
-        logger.error("Control commit failed during gateway 'reconfigure'")
-        return jsonify(message="Could not commit controls - {}".format(err)), 500
-    config.refresh()
-    return jsonify(message="Target reconfigured."), 200
+    else:
+        # DELETE target request
+        config.refresh()
+        target_config = config.config['targets'][target_iqn]
+        hostnames = target_config['portals'].keys()
+        if not hostnames:
```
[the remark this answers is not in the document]
ricardoasmarques: You are right, I can remove this.

ricardoasmarques:
@dillaman Fixed the issues you reported

ricardoasmarques:
@mikechristie Thanks, I dropped those outdated comments

ceph_iscsi_config/client.py
```diff
@@ -181,7 +182,8 @@ def define_client(self):
 
         # NB. this will check all tpg's for a matching iqn
         for client in r.node_acls:
-            if client.node_wwn == self.iqn:
+            target = client.parent_tpg.parent_target
+            if client.node_wwn == self.iqn and target.wwn == self.target_iqn:
```
mikechristie: Instead of looping over all the clients, you can just limit it to the specific target since you added it above. Something like

```
target = Target(iscsi, self.target_iqn, 'lookup')
for each tpg in target.tpgs:
     for client in tpg.node_acls:
```

ceph_iscsi_config/client.py
```diff
@@ -199,7 +201,7 @@ def define_client(self):
         # The configuration only has one active tpg, so pick that one for any
         # acl definitions
         for tpg in r.tpgs:
-            if tpg.enable:
+            if tpg.enable and tpg.parent_target.wwn == self.target_iqn:
```
mikechristie: You can just detect this in the "for each tpg in target.tpgs:" loop too.

ceph_iscsi_config/common.py
```diff
@@ -154,6 +154,59 @@ def _get_ceph_config(self):
...
+            for _, group in self.config['groups'].items():
+                group.pop('created', None)
+                group.pop('updated', None)
+            target = {
+                'disks': list(self.config['disks'].keys()),
+                'clients': self.config['clients'],
+                'portals': portals,
+                'groups': self.config['groups'],
+                'controls': self.config['controls'],
+                'ip_list': self.config['gateways']['ip_list']
```
mikechristie: I think the naming of ip_list and portals is off which is not your fault but we can clean it up here. In the iSCSI spec a target side portal is the IP address and port that can be used for iscsi traffic (we will ignore port since we always use the default 3260).

In functions like in ceph_iscsi_config/gateway.py and define_gateways  we use ip_list to create TPGs which are used for iscsi traffic, so matches the iscsi spec definition, but in functions like requires_restricted_auth we use it like a gateway IP list.

In the new config definition 'portals' looks like the old gateway but we seem to use the info in it both as a gateway definition and also a iSCSI portal definition.

Are you trying to prepare this for multiple active optimized portals per gateway?

ceph_iscsi_config/gateway.py
```diff
@@ -408,7 +410,9 @@ def map_luns(self, config):
             for tpg in self.tpg_list:
                 self.logger.debug("processing tpg{}".format(tpg.tag))
 
-                if not self.lun_mapped(tpg, stg_object):
+                target_config = config.config["targets"][self.iqn]
+                if not self.lun_mapped(tpg, stg_object) \
+                        and stg_object.name in target_config['disks']:
```
mikechristie: I think you want to change the top loop. Instead of looping over every SO just start disks defined for the specific target.

ceph_iscsi_config/gateway.py
```diff
@@ -194,15 +193,16 @@ def enable_active_tpg(self, config):
...
         """
         Remove the target definition form LIO
         :return: None
         """
         # check that there aren't any disks or clients in the configuration
         lio_root = root.RTSRoot()
 
-        disk_count = len([disk for disk in lio_root.storage_objects])
+        disk_count = len([disk for disk in lio_root.storage_objects
+                          if disk.name in target_disks])
```
mikechristie: I think here too just loop over the disks specific to the target being cleared. I think you can just merge it into the tpg loop and check for luns under the tpg since that would detect if there are disks possibly being accessed through the target still.

ceph_iscsi_config/client.py
ricardoasmarques: Looks better now, thanks. Fixed.

ceph_iscsi_config/client.py
ricardoasmarques: Fixed

ceph_iscsi_config/gateway.py
ricardoasmarques: Fixed

ceph_iscsi_config/gateway.py
ricardoasmarques: Fixed

ceph_iscsi_config/common.py
ricardoasmarques: @mikechristie No, I'm not preparing multiple active optimized portals per gateway, yet.

I'm willing to improve this, but I'll need more help / input here.

This is what we have at the moment:
````
"targets": {
            "clients": {},
            "controls": {},
            "created": "2018/12/20 13:47:49",
            "disks": [],
            "groups": {},
            "ip_list": [
                "192.168.100.201",
                "192.168.100.202"
            ],
            "portals": {
                "node1": {
                    "gateway_ip_list": [
                        "192.168.100.201",
                        "192.168.100.202"
                    ],
                    "inactive_portal_ips": [
                        "192.168.100.202"
                    ],
                    "portal_ip_address": "192.168.100.201",
                    "tpgs": 2
                },
                "node2": {
                    "gateway_ip_list": [
                        "192.168.100.201",
                        "192.168.100.202"
                    ],
                    "inactive_portal_ips": [
                        "192.168.100.201"
                    ],
                    "portal_ip_address": "192.168.100.202",
                    "tpgs": 2
                }
            },
            "updated": "2018/12/20 13:49:05"
        }
    },
````

What would you like to see renamed / moved in the config file?

Maybe we should rename `targets/<iqn>/portals` to `targets/<iqn>/gateways`, and `targets/<iqn>/ip_list` to `targets/<iqn>/portal_ip_lis`, and end up with something like:

````
"targets": {
            "clients": {},
            "controls": {},
            "created": "2018/12/20 13:47:49",
            "disks": [],
            "groups": {},
            "portal_ip_list": [
                "192.168.100.201",
                "192.168.100.202"
            ],
            "gateways": {
                "node1": {
                    "gateway_ip_list": [
                        "192.168.100.201",
                        "192.168.100.202"
                    ],
                    "inactive_portal_ips": [
                        "192.168.100.202"
                    ],
                    "portal_ip_address": "192.168.100.201",
                    "tpgs": 2
                },
                "node2": {
                    "gateway_ip_list": [
                        "192.168.100.201",
                        "192.168.100.202"
                    ],
                    "inactive_portal_ips": [
                        "192.168.100.201"
                    ],
                    "portal_ip_address": "192.168.100.202",
                    "tpgs": 2
                }
            },
            "updated": "2018/12/20 13:49:05"
        }
    },
````

Is this enought? Thoughs?

With regards to the `requires_restricted_auth` function, maybe we should use `targets/<iqn>/gateways/<host>/gateway_ip_list` instead of `targets/<iqn>/portal_ip_list`.

Does this changes address your comment?

**Comment by ricardoasmarques** [comment]
I've updated this PR with a new commit that will allow the user to create and delete disk even if there is no gateways defined.

This will prevent user from ending up in a situation where he cannot delete the disk because target(s) were deleted (and consequently the respective gateways).

File: ceph_iscsi_config/lun.py [hunk 16]
```diff
@@ -1016,7 +1013,18 @@ def define_luns(logger, config, gateway):
...
+                                disk_config = config.config['disks'][disk_key]
+                                is_allocating_host = local_gw == disk_config['allocating_host']
+                                is_lun_mapped = False
+                                for _, target_config in config.config['targets'].items():
+                                    if local_gw in target_config['portals'] \
+                                            and disk_key in target_config['disks']:
+                                        is_lun_mapped = True
+                                        break
+
+                                if is_allocating_host or is_lun_mapped:
```
dillaman: Does this prevent the creation of an associated storage object if the LUN isn't mapped to a target or the current node isn't the original node to create the disk? If so, I like the idea of deferring the creation of the backing storage object. However, I wonder if we could also eliminate the need for the allocating host and always allow the local ```rbd-target-api``` to delete the disk upon request (since the storage object should be removed upon the removal of the last mapping).

ricardoasmarques:
... and pushed another commit that will allow disk deletion from configuration / LIO, without removing the RBD image.

By default, RBD image will be deleted (to preserve old behavior), but honestly, I don't agree with this default, what about changing the default?

gwcli/storage.py
```diff
@@ -402,7 +402,7 @@ def ui_command_info(self, image_id):
         else:
             self.logger.error("disk name provided does not exist")
 
-    def ui_command_delete(self, image_id):
+    def ui_command_delete(self, image_id, delete_rbd=None):
```
dillaman: Nit: since you would need to pass the optional to keep the image, perhaps it should renamed to something like ```keep_image``` or ```preserve_image```? Also, can you add it to the help text below?

dillaman:
> By default, RBD image will be deleted (to preserve old behavior), but honestly, I don't agree with this default, what about changing the default?

... or perhaps we add a new ```detach``` disk command to emphasis that it's just detaching the disk from LIO? Perhaps the same logic could apply for the ```create``` and a new ```attach``` command.

ceph_iscsi_config/lun.py
ricardoasmarques: > Does this prevent the creation of an associated storage object if the LUN isn't mapped to a target or the current node isn't the original node to create the disk?

Yes, this is exactly what this is doing

> However, I wonder if we could also eliminate the need for the allocating host...

I think that we always have to allocate one backing storage object for each disk declared in the configuration file to prevent the user from deleting the RBD image directly from ceph and end up with an inconsistent `gateway.conf` file that contains disks pointing to RBD images that no longer exist. Am I wrong?

ricardoasmarques:
Good point, adding new commands will improve the usability.

For now, I've implemented the `detach` command. Will take care of the `attach` command latter.

ceph_iscsi_config/lun.py
dillaman: Good point. My concern was about the potential awkwardness of telling a ```gwcli``` user that they had to go back to the original host they used to create the image to remove it. If we just don't define the storage object and someone deletes it, the ```LUN.define_luns``` method could be tweaked to ignore ```rbd.ImageNotFound``` in the case where the disk doesn't have any mappings -- and the UI could display the disk attributes as a red ```NOT FOUND``` in the UI tree.

ricardoasmarques:
@dillaman pushed a commit with the `attach` command

ceph_iscsi_config/lun.py
ricardoasmarques: @dillaman I've pushed a commit that will only create LIO storage objects when needed, so the user will be able to delete unmapped disks from any host.

ricardoasmarques:
@dillaman all your comments should be addressed now.
I will be on vacation next week, so expect some delay in my responses.

ricardoasmarques:
FYI, I've pushed a new commit to this PR that adds the support for **mutual chap authentication**.

ceph_iscsi_config/common.py
ricardoasmarques: @mikechristie  Looks like we are missing something to make the separation between `portals` and `gateways`...

The only command that allows me to add an IP address is the following:

```
cd iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw/gateways
create node1 192.168.100.201
```

And the intention of this command is to create a `portal`, not a `gateway` right?


Maybe we should have a dedicated command do create `gateways` in the global configuraton, something like:

```
cd /gateways
create node1 <some_ip>
```

and only after creating the gateway ( e.g. "node1"), we should be able to create the corresponding portals inside the target:

```
cd iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw/portals
create node1 <same_or_other_ip>
```

Does this makes sense? Does it solve the problem?

dillaman: I think the original intent was to simplify the admin actions (consolidate the addition of a gateway and portal into a single step since it probably covers the vast majority of use-cases). Perhaps we can just add new sub-commands to the gateway node to add/remove additional portal address (i.e. first one is implicit)?

ricardoasmarques: @mikechristie  @dillaman Imagine a scenario where we want to create a gateway with a public IP, that should be used to **access the REST API**, but should **NOT** be used for **iSCSI traffic**.

I think that the only way to set this up is by creating an explicit separation between gateways and portals.

For _backward compatibility_, we can automatically create the gateway if it doesn't exist.

What about something like this?

```
/> ls /
...
- gateways
- iscsi-targets
  - iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw
    - portals
...

/> cd /iscsi-targets/iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw/portals

/iscsi-targets/iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw/portals> create node1 192.168.100.201

/iscsi-targets/iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw/portals> ls /
...
- gateways
  - node1 (192.168.100.201)
- iscsi-targets
  - iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw
    - portals
      - node1 (192.168.100.201)
...

/iscsi-targets/iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw/portals> create node1 192.168.101.1

/iscsi-targets/iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw/portals> ls /
...
- gateways
  - node1 (192.168.100.201)
- iscsi-targets
  - iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw
    - portals
      - node1 (192.168.100.201, 192.168.101.1)
...

/> cd /gateways

/gateways> create node2 192.168.100.202

/gateways> ls /
...
- gateways
  - node1 (192.168.100.201)
  - node2 (192.168.100.202)
- iscsi-targets
  - iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw
    - portals
      - node1 (192.168.100.201, 192.168.101.1)
...
```

ricardoasmarques:
FYI, I've pushed a new commit that adds the support for **discovery authentication (chap and chap_mutual)**.

ricardoasmarques:
I've just discovered that multiple IPs on the same host is already supported by using a list of IPs in the gateway creation:

`/iscsi-target...-igw/gateways> create node2 10.0.2.15,192.168.100.202`

This means that this PR is now completed, and ready for a new review.

dillaman:
@ricardoasmarques Couple things:

1) Starting from a 2 node blank configuration, I added a disk and and tied it to a target but the disk was never added to the backing store on the second node. As a result, attempting to map the LUN to a host resulted in a failure when it attempted to set up the second node.
2) Need feedback from @mikechristie: can CHAP settings (enabled/disabled) be different between targets on the same node? Previously, I knew that the CHAP enabled/disabled state needed to be consistent across all hosts under a given target.

Otherwise, lgtm

ricardoasmarques:
> 1. Starting from a 2 node blank configuration, I added a disk and and tied it to a target but the disk was never added to the backing store on the second node. As a result, attempting to map the LUN to a host resulted in a failure when it attempted to set up the second node.

This should be better now because, since my last commit, I'm using the same "pattern" used for other operations, where the config file will only be committed if it is applied with success on the local_gw (which is the first one where the config is applied).
This means that if something fails on local_gw while attaching a disk to a target, the config file will not be committed.
However, if the config is applied with success on `local_gw` and fails on other gateway, the config is commited and we end up with an inconsistency between the config file and the LIO, but this is not a regression and I guess this is a known issue because it's exactly the same behaviour that we have on master. I think that this last scenario should be fixed on a separate PR.

rbd-target-api.py
```diff
@@ -582,35 +648,192 @@ def _gateway(gateway_name=None):
...
+        # process other gateways first
+        gateways.append(local_gw)
+        resp_text, resp_code = call_api(gateways, '_targetlun',
+                                        '{}'.format(target_iqn),
+                                        http_method='delete',
+                                        api_vars=api_vars)
+        if resp_code != 200:
+            return jsonify(message="Delete target LUN mapping failed - "
+                                   "{}".format(resp_text)), resp_code
+
```
dillaman: I think you need a ```config.refresh()``` here. With a blank configuration, the first time I add a disk it gets sent to the second node. The gwcli queries the config from the local node after adding the target LUN but fails since the config from the local node is missing the ```owner``` key in the disk so the gwcli crashes. I added it to my local install, retested, and it worked.

dillaman: ... or perhaps the "config" API entry point should always attempt a refresh before returning the config to ensure it's always up-to-date regardless of where changes were made.

ricardoasmarques: Good spotting. I haven't detected this because, on my system, the `config_watcher` updated the config before the gwcli request, so everything worked without the explicit config.refresh.

Should be fixed now, thanks.

PR #10, opened 2018-12-12
Disk size is optional if image already exists

ricardoasmarques:
User should be able to create a disk using an existing RBD image, but in this case, it doesn't make sense to force the user to specify the disk size.

This PR will make the `size` parameter optional when image already exists.

Note that in Ceph Manager Dashboard we will use images that were previously created on the dedicated `Block > Images` page.

### DIFF
--- a/gwcli/client.py
+++ b/gwcli/client.py
@@ -475,10 +475,6 @@ class Client(UINode):
             if action == 'add':
                 ui_root = self.get_ui_root()
                 ui_disks = ui_root.disks
-                if not size:
-                    self.logger.error("To auto-define the disk to the client"
-                                      " you must provide a disk size")
-                    return
 
                 # a disk given here would be of the form pool.image
                 pool, image = disk.split('.')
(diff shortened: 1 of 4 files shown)


gwcli/client.py
```diff
@@ -475,10 +475,6 @@ def ui_command_disk(self, action='add', disk=None, size=None):
             if action == 'add':
                 ui_root = self.get_ui_root()
                 ui_disks = ui_root.disks
-                if not size:
```
dillaman: Can the ```size``` parameter (from the CLI action) be dropped now?

ricardoasmarques: The `size` parameter is still needed if the RBD image does not exist yet, and we want to create it using `ceph-iscsi` (this PR is just changing the parameter from mandatory to optional).

Dropping this parameter would require the drop of the functionality that creates RBD images directly from the `ceph-iscsi`.

dillaman: OK, I guess I didn't realize that you could create brand new disks from under the "host/XYZ" UI tree

PR #12, opened 2019-01-15
Use normalized iscsi names

mikechristie:
The iSCSI spec defines the initiator/target names as case insensitive and rtslib will convert capital letters to lower case. If the user passes in upper case we will store that non-normalized format in gateway.conf then later try to match that with the rtslib/kernel value. So we can end up creating clients/targets with capital letters but not remove them.

This patch series has us normalize and use that name at the gwcli/user level, API layer and lib/ceph-iscsi-config layers, so if it is user input or an app creating these objects we will always use the value rtslib and the kernel uses.

### DIFF
--- a/gwcli/client.py
+++ b/gwcli/client.py
@@ -2,13 +2,14 @@
 
 from gwcli.node import UIGroup, UINode
 
-from gwcli.utils import response_message, valid_iqn, APIRequest
+from gwcli.utils import response_message, APIRequest
 
 from ceph_iscsi_config.client import CHAP
 import ceph_iscsi_config.settings as settings
 from ceph_iscsi_config.utils import human_size
 
 import rtslib_fb.root as root
+from rtslib_fb.utils import normalize_wwn, RTSLibError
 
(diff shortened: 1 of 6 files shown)


gwcli/client.py
```diff
@@ -2,13 +2,14 @@
 
 from gwcli.node import UIGroup, UINode
 
-from gwcli.utils import response_message, valid_iqn, APIRequest
```
dillaman: Nit: can we delete the ```valid_iqn``` function if it's no longer used?

mikechristie: It's still used in valid_client() to check if the name is in correct format.

I was thinking I could just remove that check since it is normalized before valid_client() is called but I was thinking maybe other callers might want to use valid_client to check still.

dillaman: Can that method be updated to use the same ```normalize_wwn``` method?

mikechristie: Yeah, I see what you mean now. Done.

mikechristie:
Repushed with added patch to drop valid_iqn.

PR #14, opened 2019-01-19
Handle image with snapshot deletion exception

mikechristie:
Instead relying on the generic exception handler, catch
ImageHasSnapshots and pass a informative error message to the user.

### DIFF
--- a/ceph_iscsi_config/lun.py
+++ b/ceph_iscsi_config/lun.py
@@ -71,6 +71,8 @@ class RBDDev(object):
         """
 
         rbd_deleted = False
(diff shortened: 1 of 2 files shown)


mikechristie:
Oh shoot. Just adding myself a note about why we fixed this twice in the 2 repos. I just forgot about Jason fixing this in ceph-iscsi-config a couple months ago:

https://github.com/ceph/ceph-iscsi-config/pull/94

I think we want Jason's fix instead of mine because it is done earlier before we start to tear stuff down.

PR #17, opened 2019-01-23
Optimize the speed of creating disks

swimfish09:
[https://github.com/ceph/ceph-iscsi-cli/issues/143](url)

before:
image
now:
image

### DIFF
--- a/ceph_iscsi_config/gateway.py
+++ b/ceph_iscsi_config/gateway.py
@@ -323,6 +323,7 @@ class GWTarget(GWObject):
             self.target = [tgt for tgt in lio_root.targets
                            if tgt.wwn == self.iqn][0]
 
+            self.tpg_list = []
             # there could/should be multiple tpg's for the target
             for tpg in self.target.tpgs:
                 self.tpg_list.append(tpg)
@@ -418,8 +419,8 @@ class GWTarget(GWObject):
         for tpg in self.tpg_list:
 
             for l in tpg.luns:
-                tpg_stg_object_dict[str(tpg.tag) + "-"
-                 + l.storage_object.name] = l.storage_object.name
+                key = str(tpg.tag) + "-" + l.storage_object.name
+                tpg_stg_object_dict[key] = l.storage_object.name
 
         # process each storage object added to the gateway, and map to the tpg
         for stg_object in target_stg_object:
--- a/ceph_iscsi_config/gateway.py
+++ b/ceph_iscsi_config/gateway.py
@@ -412,21 +412,22 @@ class GWTarget(GWObject):
         target_config = config.config["targets"][self.iqn]
         target_stg_object = [stg_object for stg_object in lio_root.storage_objects
                              if stg_object.name in target_config['disks']]
-        
+
         # a dict, key with tpg and storage object name
         tpg_stg_object_dict = {}
         for tpg in self.tpg_list:
 
-             for l in tpg.luns:
-                tpg_stg_object_dict[str(tpg.tag) + '-' + l.storage_object.name] = l.storage_object.name
-        
+            for l in tpg.luns:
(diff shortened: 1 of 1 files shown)


swimfish09:
@mikechristie

ceph_iscsi_config/gateway.py
```diff
@@ -412,14 +412,21 @@ def map_luns(self, config):
...
+                tpg_stg_object_dict[str(tpg.tag) + '-' + l.storage_object.name] = l.storage_object.name
+        
         # process each storage object added to the gateway, and map to the tpg
         for stg_object in target_stg_object:
 
             for tpg in self.tpg_list:
                 self.logger.debug("processing tpg{}".format(tpg.tag))
 
-                if not self.lun_mapped(tpg, stg_object):
+                if not tpg_stg_object_dict.has_key(str(tpg.tag) + "-" + stg_object.name):
```
mikechristie: Did you see my comment:

https://github.com/ceph/ceph-iscsi-config/pull/100#discussion_r250003141

LUN() does a exists check for us, so there is no need to create the dict. Did my suggestion in that comment not perform as well as the dict based approach above?

swimfish09: yeah, i see. but when catch a exception, configuration will be cleaned and the program exit.

swimfish09: image

dillaman:
@swimfish09 Can you address the tox errors?

ceph_iscsi_config/gateway.py
mikechristie: It's not clear to me how you hit that. Won't the code:

```
        except RTSLibError as err:
            if "already exists in configFS" not in str(err):
                self.logger.error("LUN mapping failed: {}".format(err))
                self.error = True
                self.error_msg = str(err)
                return

           #Already created. Ignore and loop to the next tpg.
           continue
```

Catch the exception so it does not get escalated up to exception_handler()?

swimfish09: use this code, rbd-target-gw.log with no error 
image

swimfish09:
> @swimfish09 Can you address the tox errors?
rbd-target-gw.log with no error

dillaman:
@swimfish09 These errors from GitHub [1]

[1] https://jenkins.ceph.com/job/ceph-iscsi-tox/49/console

swimfish09:
Add a dict to check lun is or not mapped will cause errors, the same lun will be mapped twice 
image

When i add code print self.tpg_list to method map_luns, i found self.tpg_list contains repeat tpg. 
Is this correct that self.tpg_list contains repeat tpg?
image
image

swimfish09:
Is this is a error? tpg_list contains repeat tpg
image

image

dillaman:
@mikechristie look good to you?

mikechristie:
I am checking my own review comment to see why it does not work.

mikechristie:
For some reason you cannot attach a file in a nested comment, so I am replying here.

Attached is a patch that works for me and avoids the building of the tpg/lun cache. Exceptions are handled as expected, so if you are hitting issues remove the generic handling in rbd-target-api in  unhandled_exception(), so we can get a trace and double check it is coming from this path.

[skipscan.txt](https://github.com/ceph/ceph-iscsi/files/2882426/skipscan.txt)

mikechristie:
Doh. I posted that then realized for disk creation we do not even need to loop over all the storage_objects too since we just want to add that one disk. Let me open a new PR, because we also do not need to do lun.map_lun->define_luns too (we just need to map the one lun).

mikechristie:
I integrated the idea of your patch to remove the excess lun dir scans into my patchset here:

https://github.com/ceph/ceph-iscsi/pull/34

but based it on my patches. I'm gonna close this PR as a result.

PR #18, opened 2019-01-23
Fix race condition after target creation

ricardoasmarques:
How to reproduce (starting with a clean configuration):

1. API call to create target on node1
2. API call to create gateway node1 192.182.100.201
3. API call to create gateway node2 192.182.100.202

Sometimes it fails on step `3` because the newly created target doesn't exist in the node2 config yet


ricardoasmarques:
@mikechristie moving it to the “call_api” will not fix the issue that I’m trying to solve with this PR because, in the scenario I’ve described, we start with a clean configuration, and node2 is not a known gateway in the moment we create the target (step1), so target creation will not be executed in node2 anyway.

PR #23, opened 2019-02-08
Adds infrastructure for multiple backstores

ricardoasmarques:
**Motivation**
In _openSUSE_, we have 2 different "backstores" so we have to patch `ceph-iscsi` in order to support the `kernel-lio-rbd` in addition to the already supported `tcmu-runner`.

**Proposed Solution**
With this PR, `ceph-iscsi` will continue to only support `tcmu-runner` but the code (and `gateway.conf`) will be prepared to easily add support for a different backstore.
This proposed change will not only simplify the _openSUSE_ patch, but also be useful if, in the future, `ceph-iscsi` has to support any additional backstore.

### DIFF
--- a/ceph_iscsi_config/backstore.py
+++ b/ceph_iscsi_config/backstore.py
@@ -1,0 +1,4 @@
+#!/usr/bin/env python
+
+
+USER_RBD = 'user:rbd'
--- a/ceph_iscsi_config/common.py
+++ b/ceph_iscsi_config/common.py
@@ -5,6 +5,7 @@ import time
 import json
 import traceback
 
+from ceph_iscsi_config.backstore import USER_RBD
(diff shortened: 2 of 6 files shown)


ricardoasmarques:
@dillaman I'm now using a constant for the backstore name

ceph_iscsi_config/backstore.py
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env python
+
+
+class Backstore:
```
dillaman: Nit: probably no need for the class since its full name is "ceph_iscsi_config.backstore.USER_RBD"

ricardoasmarques: Fixed

PR #24, opened 2019-02-11
Adds support for disabling ACL authentication

ricardoasmarques:
Initiator name based ACL authentication can now be disabled using the `auth noacl` command (if the target doesn't contain clients).
ACL authentication will be automatically re-enabled if one client is added to the target.

### DIFF
--- a/ceph_iscsi_config/gateway.py
+++ b/ceph_iscsi_config/gateway.py
@@ -231,6 +231,14 @@ class GWTarget(GWObject):
             self.error = True
             self.error_msg = "Unable to delete target - {}".format(err)
 
+    def update_acl(self, config):
+        target_config = config.config["targets"][self.iqn]
+        for tpg in self.tpg_list:
+            if target_config['acl_enabled']:
+                tpg.set_attribute('generate_node_acls', 0)
+            else:
+                tpg.set_attribute('generate_node_acls', 1)
+
     def create_tpg(self, ip):
 
         try:
@@ -550,6 +558,8 @@ class GWTarget(GWObject):
 
                 self.map_luns(config)
 
+                self.update_acl(config)
+
             else:
                 self.error = True
                 self.error_msg = ("Attempted to map to a gateway '{}' that "
@@ -570,6 +580,7 @@ class GWTarget(GWObject):
                 seed_target = {
                     'disks': [],
                     'clients': {},
+                    'acl_enabled': True,
                     'portals': {},
                     'groups': {},
                     'controls': {}
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -1486,6 +1486,79 @@ def _discoveryauth():
     return jsonify(message='OK'), 200
 
 
+@app.route('/api/targetauth/<target_iqn>', methods=['PUT'])
+@requires_restricted_auth
+def targetauth(target_iqn=None):
+    """
+    Coordinate the gen-acls across each gateway node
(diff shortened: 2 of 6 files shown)


ceph_iscsi_config/gateway.py
```diff
@@ -231,6 +231,14 @@ def clear_config(self):
             self.error = True
             self.error_msg = "Unable to delete target - {}".format(err)
 
+    def set_noacl(self, config):
+        target_config = config.config["targets"][self.iqn]
+        for tpg in self.tpg_list:
+            if target_config['noacl']:
+                tpg.set_attribute('generate_node_acls', 1)
```
dillaman: I'm not familiar w/ the configfs layout, but setting ```generate_node_acls``` to 1 will disable ACLs? Seems backwards to me.

rbd-target-api.py
```diff
@@ -1468,6 +1468,70 @@ def _discoveryauth():
     return jsonify(message='OK'), 200
 
 
+@app.route('/api/noacl/<target_iqn>', methods=['PUT'])
```
dillaman: Nit: the ```noacl``` endpoint just seems odd to me. How about something like ```/api/targetauth/<target_iqn>``` with a parameter to disable ACLs?

ceph_iscsi_config/gateway.py
ricardoasmarques: Yes, the attribute name may be confusing, but it disabled the ACL authentication - https://wiki.alpinelinux.org/wiki/Linux_iSCSI_Target_(TCM)

rbd-target-api.py
ricardoasmarques: `noacl` endpoint was renamed to `targetauth` with an additional parameter to disable ACL.

Additionally I've moved the `noacl` field in the configuration file from `/targets/<target_iqn>/noacl` to `/targets/<target_iqn>/auth/noacl`.

ricardoasmarques:
@dillaman your comments were addressed.

ricardoasmarques:
I've pushed a new commit that will allow the user to re-enable the ACL authentication by performing the following command: `auth acl`.

ricardoasmarques:
Rebased

mikechristie:
My only comment/question is that I think I messed up on the nochap naming. I am not really sure why I did that. I think it makes more sense to do:

auth_type = initiator_name | chap | none

Unfortunately nochap is already user facing, but do we want to depreciate that (there are not many users so it would only need to be for a release) and also not carry it over with noacl? If you guys thought the nochap/acl naming was weird then, it would be ok with me.

dillaman:
@mikechristie I am definitely confused by it, so if we can make it clearer, it would be better.

ricardoasmarques:
I'll work on this

ricardoasmarques:
@mikechristie @dillaman I've now pushed a new proposal which I believe is less confusing.

This PR will now add 2 new commands:

- `auth disable_acl` - Disable initiator name based ACL authentication
- `auth enable_acl` - Enable initiator name based ACL authentication

From the configuration point of view, this PR adds one new field to persist if the ACL is enabled:

```
{
  "targets": {
      "acl_enabled": True,
      ...
    }
  }
}
```

dillaman:
lgtm -- @mikechristie any comments/objections?

mikechristie:
LGTM.

rjfd:
@dillaman this looks good to be merged, or are there any concerns still?

PR #29, opened 2019-02-18
 Support dots in pool and image names

ricardoasmarques:
Using a `.` to separate the `pool name` and the `image name` is very restrictive since it will not allow us to export pools/images that contains a `dot` in the name (which is a valid name).

In this PR, we start using the "official"  `/` separator.

**Before (example)**

Command:
`attach rbd.disk_1`

Config:
```
"disks": {
            "rbd.disk_1": {
                ...
            }
        }
```

**After (example)**
Command:
`attach rbd/disk_1`

Config
```
"disks": {
            "rbd/disk_1": {
                "backstore_object_name": "rbd.disk_1",
                ...
            }
        }
```

(note that the LIO name will be the `<backstore_object_name>`, which will not contain `slashes`)


gwcli/storage.py
```diff
@@ -407,13 +406,38 @@ def ui_command_resize(self, image_id, size):
         disk = self.disk_lookup[image_id]
         disk.resize(size)
 
+    def ui_command_snapshot(self, image_id, action, name):
```
dillaman: Should this be part of a different PR?

gwcli/storage.py
```diff
@@ -882,53 +909,6 @@ def _update_pool(self):
             # update the pool commit numbers
             pool._calc_overcommit()
 
-    def ui_command_resize(self, size):
```
dillaman: Why remove all of these?

ricardoasmarques: Because disk name will now contain slashes:

```
o- disks ..................................................... [92G, Disks: 2]
  | o- rbd/disk_1 ............................................ [disk_1 (90G)]
  | o- rbd/disk_2 ............................................ [disk_2 (2G)]
```
we are no longer able to do `cd /disks/<disk_id>`, so I've moved all disk commands to the `/disks` level.

E.g., you will now have to do 
```
cd /disks
resize rbd/disk_1 30G
```

instead of 
```
cd /disks/rbd/disk_1
resize 30G
```

dillaman: Now that you bring this up, that was exactly the reason why the dots were used. Can we just add a new sub-level to the UI for the pool? or use the backstore name in the UI?

gwcli/storage.py
ricardoasmarques: Have to be done in this PR because it's a limitation of adding the support for `dots` in the name (as explained in my previous comment, you cannot do `cd /disks/<pool>/<image>`. So we have to move this command from `/disks/<disk_id>` to `/disks`.

gwcli/storage.py
ricardoasmarques: I'll add a new sub-level for the pool.

ricardoasmarques:
@dillaman a new pool sub-level has been added to the disks section:

screenshot from 2019-02-25 17-53-57

mikechristie:
Don't forget to update the man page gwcli.8.

ricardoasmarques:
@mikechristie Done

dillaman:
@ricardoasmarques rebase please

ricardoasmarques:
@dillaman This PR is now rebased

ceph_iscsi_config/lun.py
```diff
@@ -1060,6 +1063,29 @@ def get_owner(gateways, portals):
...
+        image name, and existing storage object names to avoid conflicts
+        :param pool: pool name
+        :param image: image name
+        :param disks_config: disks configuration from `gateway.conf`
+        :return: the backstore storage object name to be used
+        """
+        base_name = '{}.{}'.format(pool, image)
+        candidate = base_name
+        counter = 0
+        while LUN._backstore_object_name_exists(disks_config, candidate):
```
mikechristie: Hey, what was this for? I'm a little confused because I think we have multiple ways to create multiple disks but I do not think we hit this code path. We seem to have:

1. gwcli disk creation has a "count" argument. In this case create_disk seems to loop and allocate a index value to add to the image name. For this, only gwcli sees count and it is not passed to #2.

2. The REST call "/api/disk/<pool>/<image>" has a "count" argument. disk() seems to have a loop where it also allocates a index value to add to the image name, so we would not hit this code path that way.

3. In rbd-target-api:disk() it seems that when it calls valid_disk there is a check for if there is an existing image name in the pool, so I do not think we will loop here since we would fail in the valid_disk call already.

Maybe did you mean to merge all that stuff too?

ricardoasmarques: @mikechristie The purpose of this code is just to avoid storage object names conflicts, it's not related to multiple disks creation.

I've improved the documentation of this method to include an example that may be useful.

In summary, the storage object name will use a `.` as a separator for `pool name` and `image name`, which may cause conflicts in scenarios like this one:

  - Add disk `a.b/c` will create the storage object `a.b.c`
  - Add disk `a/b.c` will create the storage object `a.b.c.1`

Without name conflict resolution we will have an error in the second disk creation, so I think there's nothing to change in this code.

mikechristie: Makes sense. LGTM.

PR #31, opened 2019-02-20
rpm packaging fixes

ktdreyer:
Drop tabs for spaces, update the `License` to `GPLv3+`, use the `%autosetup` macro, and add Python 3 support for Fedora and RHEL 8.

### DIFF
--- a/ceph-iscsi.spec
+++ b/ceph-iscsi.spec
@@ -75,7 +75,7 @@ performance statistics, supporting monitoring and visualisation tools like
 Grafana.
 
 %prep
-%setup -q
+%autosetup -p1
 
 %build
 %if 0%{?with_python2}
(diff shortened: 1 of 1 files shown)


ktdreyer:
I've tested building this in a RHEL 8 mock chroot and it installs at least :)

dillaman:
@ricardoasmarques @rjfd The current spec doesn't have any openSUSE-specific directives, but do either of you think this PR would introduce new issues?

rjfd:
@dillaman let me give this a try tomorrow morning.

ricardoasmarques:
@dillaman I haven't tested this PR, but I don't think this will introduce any issue.

In openSUSE we are using python 3 and the changes we made on the spec file are similar to the ones in this PR, so lgtm.

ricardoasmarques:
@smithfarm just FYI.

rjfd:
@dillaman this PR is good to go.

Regarding openSUSE, there are some differences on the way python packages are built. If you don't oppose, I would like to open a PR that adds the opensuse specific parts within some conditionals (as we already do with ceph.spec). This would allows us to drop a downstream version of the spec file.

dillaman:
@rjfd Sounds good to me, thanks.

ktdreyer:
good idea @rjfd

PR #32, opened 2019-02-21
Makes pool name configurable

ricardoasmarques:
With this PR, the `pool name` used to store the `gateway.conf` can be configured in `/etc/ceph/iscsi-gateway.cfg` file (defaults to `rbd`).


ricardoasmarques:
@dillaman all occurrences of the hardcoded 'rbd' string are now replaced by the corresponding config setting.

ricardoasmarques:
@rjfd Done

PR #34, opened 2019-02-23
Lun/target map and lookup fixes

mikechristie:
In a lot of cases we have been looping over lun and backstore/storage_object dirs when we did not need to. This is based off of @swimfish09's PR here:

https://github.com/ceph/ceph-iscsi/pull/17

but removes a couple extra loops.


ricardoasmarques:
I haven't found any issue during my manual tests

PR #35, opened 2019-02-26
spec: opensuse support

rjfd:
This PR adds some conditionals to the spec file to support the packaging in openSUSE based distros.
It also cleans up a bit the spec file, and also deletes python interpreter shebangs in non-executable python files.

### DIFF
--- a/ceph-iscsi.spec
+++ b/ceph-iscsi.spec
@@ -23,9 +23,12 @@ Name:           ceph-iscsi
 Version:        3.0
 Release:        1%{?dist}
 Group:          System/Filesystems
-Summary:        Python package providing modules for ceph iscsi gateway configuration management
-
+Summary:        Python modules for Ceph iSCSI gateway configuration management
+%if 0%{?suse_version}
+License:        GPL-3.0-or-later
+%else
 License:        GPLv3+
+%endif
 URL:            https://github.com/ceph/ceph-iscsi
 Source0:        https://github.com/ceph/ceph-iscsi/archive/%{version}/%{name}-%{version}.tar.gz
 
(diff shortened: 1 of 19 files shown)


rjfd:
@ktdreyer could you check if I didn't break rhel/centos and fedora packaging with these changes?

ceph-iscsi.spec
```diff
@@ -81,31 +90,69 @@ Grafana.
...
 install -m 0644 gwcli.8 %{buildroot}%{_mandir}/man8/
 gzip %{buildroot}%{_mandir}/man8/gwcli.8
+%if 0%{?suse_version}
+mkdir -p %{buildroot}%{_unitdir}/rbd-target-gw.service.d
+install -m 0644 .%{_sysconfdir}/systemd/system/rbd-target-gw.service.d/dependencies.conf %{buildroot}%{_unitdir}/rbd-target-gw.service.d/
+mkdir -p %{buildroot}%{_sbindir}
+ln -s service %{buildroot}%{_sbindir}/rcrbd-target-gw
+ln -s service %{buildroot}%{_sbindir}/rcrbd-target-api
+%else
+mkdir -p %{buildroot}%{_sysconfdir}/systemd/system/rbd-target-gw.service.d
```
ktdreyer: Oh, yes this is a general bug in our systemd packaging here. We should install to `%{_unitdir}` on all platforms and not write files into `/etc/systemd`. Are you ok making this change for everything?

We'll still need the `suse_version` conditional here (and below in `%files`) for the rc* scripts.

ceph_iscsi_config/backstore.py
```diff
@@ -1,4 +1,2 @@
-#!/usr/bin/env python
-
```
smithfarm: drop this line so the file doesn't start with a blank line?

ceph-iscsi.spec
rjfd: Sure, I'll fix this.

rjfd:
@ktdreyer @smithfarm addressed your comments.

PR #36, opened 2019-02-27
Add support for specifying cluster client name

rjfd:
Currently ceph-iscsi services will always try to connect to the cluster as `client.admin` client. This PR allows to change that.

### DIFF
--- a/ceph_iscsi_config/settings.py
+++ b/ceph_iscsi_config/settings.py
@@ -94,8 +94,8 @@ class Settings(object):
         return v
 
     defaults = {"cluster_name": "ceph",
-                "gateway_keyring": "ceph.client.admin.keyring",
                 "pool": "rbd",
+                "cluster_client_name": "client.admin",
                 "time_out": 30,
                 "api_host": "::",
                 "api_port": 5000,
--- a/ceph_iscsi_config/settings.py
+++ b/ceph_iscsi_config/settings.py
@@ -160,7 +160,7 @@ class Settings(object):
                 # We always want these values set to at least the defaults.
                 self._define_settings(Settings.target_defaults)
 
-        self.cephconf = '/etc/ceph/{}.conf'.format(self.cluster_name)
+        self.cephconf = '{}/{}.conf'.format(self.ceph_config_dir, self.cluster_name)
         if self.api_secure:
             self.api_ssl_verify = False if self.api_secure else None
 
--- a/gwcli/ceph.py
+++ b/gwcli/ceph.py
@@ -38,19 +38,19 @@ class CephGroup(UIGroup):
 
         for cluster_name in self.cluster_map.keys():
 
-            keyring = self.cluster_map[cluster_name]['keyring']
+            cluster_client_name = None
             if cluster_name == settings.config.cluster_name:
 
-                if settings.config.gateway_keyring:
-                    keyring = settings.config.gateway_keyring
-                    self.cluster_map[cluster_name]['keyring'] = keyring
+                if settings.config.cluster_client_name:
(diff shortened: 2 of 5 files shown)


ceph_iscsi_config/settings.py
```diff
@@ -94,6 +94,7 @@ def normalize(k, v):
         return v
 
     defaults = {"cluster_name": "ceph",
+                "cluster_client_name": "client.admin",
                 "gateway_keyring": "ceph.client.admin.keyring",
```
dillaman: Nit: I wonder if this should default to ```None``` / blank now so that librados can fallback to its default search paths / naming conventions for the keyring (i.e. ```<cluster-name>.<cephx name>.keyring```). Or I wonder if it just makes sense to drop the keyring since it can be specified via the Ceph configuration file anyway.

rjfd: I’m in favor of dropping it. How should we handle the upgrade? Log a warning saying that “gateway_keyring” is ignored if the Ceph-iscsi conf file still has the option set?

dillaman: I doubt anyone has ever used that override, so it should be safe to just remove.

rjfd:
@dillaman removed `gateway_keyring` in this last push.

gwcli/ceph.py
```diff
@@ -76,14 +76,20 @@ def get_clusters(self):
             name = os.path.basename(conf).split('.')[0]
             keyring = glob.glob(os.path.join(CephGroup.ceph_config_dir,
```
dillaman: I think this can be yanked as well

rjfd: I was using this to infer the client name. Are you suggesting to remove all lines below down to `local = True if name ...`?

rjfd: TBH, I don't understand the purpose of this code (detect several clusters)? Are there plans for ceph-iscsi to use RBD images from different clusters?

dillaman: To be honest, I don't understand it either. The backing API would only work with a single cluster specified in the config. It might have been a partially implemented thing for supporting mirroring, but honestly I think it can all be removed since it should be getting the mirror peer data from the pool.

rjfd: In that case I'll remove that code in this PR.

rjfd:
@dillaman I removed the multiple cluster discovery

gwcli/ceph.py
```diff
@@ -32,60 +30,15 @@ class CephGroup(UIGroup):
     default_ceph_conf = '{}/ceph.conf'.format(ceph_config_dir)
```
dillaman: Nit: unused

gwcli/ceph.py
```diff
@@ -32,60 +30,15 @@ class CephGroup(UIGroup):
...
-                              'keyring': keyring[0],
-                              'name': name,
-                              'local': local}
-
-        return clusters
+        self.logger.debug("Adding ceph cluster '{}' to the UI"
+                          .format(settings.config.cluster_name))
+        self.cluster = CephCluster(self,
+                                   settings.config.cluster_name,
+                                   "{}/{}.conf".format(settings.config.ceph_config_dir,
```
dillaman: Nit: re-use ```settings.config.cephconf```?

gwcli/ceph.py
rjfd: done

gwcli/ceph.py
rjfd: done

rjfd:
@dillaman addressed your comments, and rebased PR

dillaman:
Thanks!

PR #37, opened 2019-02-28
Support passwords with '/' (and more...)

ricardoasmarques:
Using a `/` to separate the username from the password is very restrictive and will not work if we are migrating from an environment that already has users created (which may contain 'slashes').

This PR changes `auth` commands and `discovery_auth` command to use separate parameters for username/password

The password validation will now accept '/'s.

The configuration structure changed in a way that will not depend on the '/' char in the CHAP string.

**Before (example):**

Command
```
auth chap=myusername/mypassword chap_mutual=mymusername/mympassword

# or

auth myusername/mypassword mymusername/mympassword
```

Config
```
{
  "discovery_auth": {
    "chap": ""
    "mutual_chap": ""
  }
  "targets": {
      "clients": {
        "iqn.1994-05.com.redhat:rh7-client": {
          "auth": {
            "chap": "myusername/mypassword"
            "mutual_chap": "mymusername/mympassword"
          }
        }
      }
    }
}
```

**After (example):**

Command
```
auth username=myusername password=mypassword mutual_username=mymusername mutual_password=mympassword

# or

auth myusername mypassword mymusername mympassword
```
Config
```
{
  "discovery_auth": {
    "username": ""
    "password": ""
    "mutual_username": ""
    "mutual_password": ""
  }
  "targets": {
      "clients": {
        "iqn.1994-05.com.redhat:rh7-client": {
          "auth": {
            "username": "myusername"
            "password": "mypassword"
            "mutual_username": "myusername"
            "mutual_password": "mympassword"
          }
        }
      }
    }
}
```

**Additionally, this PR includes the following changes:**

- Adds support for encrypting discovery passwords
- Do not rely on password size to check if it's encrypted
  - Instead of assuming that passwords with more than 16 chars are encrypted, we are now storing this information explicitly on `password_encryption_enabled` and `mutual_password_encryption_enabled` config fields
- It's now possible to get the `gateway.conf` file, from `/api/config` API, with passwords decrypted (by using the `decrypt_passwords` parameters)

### DIFF
--- a/ceph_iscsi_config/client.py
+++ b/ceph_iscsi_config/client.py
@@ -27,19 +27,27 @@ class GWClient(GWObject):
                 "nopin_timeout",
                 "cmdsn_depth"]
 
-    seed_metadata = {"auth": {"chap": '', "chap_mutual": ''},
+    seed_metadata = {"auth": {"username": '',
+                              "password": '',
+                              "password_encryption_enabled": False,
+                              "mutual_username": '',
+                              "mutual_password": '',
+                              "mutual_password_encryption_enabled": False},
                      "luns": {},
                      "group_name": ""
                      }
 
-    def __init__(self, logger, client_iqn, image_list, chap, chap_mutual, target_iqn):
+    def __init__(self, logger, client_iqn, image_list, username, password, mutual_username,
+                 mutual_password, target_iqn):
         """
         Instantiate an instance of an LIO client
         :param client_iqn: (str) iscsi iqn string
         :param image_list: (list) list of rbd images (pool/image) to attach
                            to this client or list of tuples (disk, lunid)
-        :param chap: (str) chap credentials in the format 'user/password'
-        :param chap_mutual: (str) chap_mutual credentials in the format 'user/password'
+        :param username: (str) chap username
+        :param password: (str) chap password
+        :param mutual_username: (str) chap mutual username
+        :param mutual_password: (str) chap mutual password
         :param target_iqn: (str) target iqn string
         :return:
         """
@@ -63,8 +71,10 @@ class GWClient(GWObject):
             else:
                 self.requested_images = image_list
 
-        self.chap = chap                        # parameters for auth
-        self.chap_mutual = chap_mutual          # parameters for mutual auth
+        self.username = username
+        self.password = password
+        self.mutual_username = mutual_username
+        self.mutual_password = mutual_password
         self.mutual = ''
         self.tpgauth = ''
         self.metadata = {}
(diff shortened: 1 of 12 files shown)


ricardoasmarques:
Pushed a new commit changing the way we check if a password is encrypted.

Instead of assuming that passwords with more than 16 chars are encrypted, we are now storing this information explicitly on `password_encryption_enabled` and `mutual_password_encryption_enabled` config fields

ceph_iscsi_config/client.py
```diff
@@ -467,12 +477,20 @@ def exists(self):
         client_list = [client.node_wwn for client in r.node_acls]
         return self.iqn in client_list
 
+    @staticmethod
+    def seed_metadata():
+        client_config = GWClient._seed_metadata
+        encryption_enabled = encryption_available()
```
dillaman: Nit: since the seed passwords are empty, shouldn't this just be hard-coded to ```False```?

ceph_iscsi_config/common.py
```diff
@@ -85,6 +89,14 @@ def __init__(self, logger, cfg_name='gateway.conf', pool=None):
             self._upgrade_config()
             self.changed = False
 
+    @staticmethod
+    def seed_config():
+        config = Config._seed_config
+        encryption_enabled = encryption_available()
```
dillaman: Nit: same comment here

ceph_iscsi_config/client.py
```diff
@@ -274,66 +289,53 @@ def try_disable_auth(self, tpg):
...
+        if username and password:
             auth_enabled = True
-        elif not chap_credentials:
-            client_username = ''
-            client_password = ''
-            chap_credentials = "/"
 
-        if '/' in chap_mutual_credentials:
-            client_mutual_username, client_mutual_password = chap_mutual_credentials.split('/', 1)
+        if mutual_username and mutual_password:
```
dillaman: Nit: if ```username``` / ```password``` are empty, it would be an error to attempt to configure mutual auth, correct? Perhaps a check that could be added to ```valid_credentials```?

gwcli/client.py
```diff
@@ -369,45 +369,40 @@ def summary(self):
 
         return ", ".join(msg), status
 
-    def set_auth(self, chap=None, chap_mutual=None):
+    def set_auth(self, username=None, password=None, mutual_username=None, mutual_password=None):
 
-        self.logger.warn("chap={}, chap_mutual={}".format(chap, chap_mutual))
+        self.logger.warn("username={}, password={}, mutual_username={}, "
```
dillaman: Nit: should this be ```self.logger.debug``` instead?

ricardoasmarques:
@dillaman your comments were addressed and PR was rebased

PR #43, opened 2019-03-12
Fix lun exists check

mikechristie:
We must convert exception to string before testing.

Note: In some versions of python we do not seem to have to do the
explicit conversion with a str() call and it is done automagically for
us. I am not sure what versions/why. It seems safe to always do the str() conversion.

This is probably why someone saw the exception get passed upwards as we were not matching the string comparison test.


dillaman:
@mikechristie Is this just the same as the second commit in #44?

mikechristie:
Yeah, I messed up. In the other PR the second patch was supposed to be a different patch. I have fixed the PRs now, so PR 44 has the correct second patch.

PR #47, opened 2019-03-13
gw: use client id when calling ceph shell command

rjfd:

### DIFF
--- a/rbd-target-gw.py
+++ b/rbd-target-gw.py
@@ -47,8 +47,10 @@ def ceph_rm_blacklist(blacklisted_ip):
     logger.info("Removing blacklisted entry for this host : "
                 "{}".format(blacklisted_ip))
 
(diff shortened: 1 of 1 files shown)


rbd-target-gw.py
```diff
@@ -47,18 +47,18 @@ def ceph_rm_blacklist(blacklisted_ip):
     logger.info("Removing blacklisted entry for this host : "
                 "{}".format(blacklisted_ip))
 
-    result = subprocess.check_output("ceph --conf {cephconf} osd blacklist rm {blacklisted_ip}".
```
dillaman: You should still keep the ```--conf``` while adding ```-n```

rjfd: that's true, I was wrongly assuming that the Ceph conf file would always be `/etc/ceph/ceph.conf`

rjfd: fixed

PR #48, opened 2019-03-13
Always use rados client name when connecting to Ceph cluster

rjfd:

### DIFF
--- a/ceph_iscsi_config/lun.py
+++ b/ceph_iscsi_config/lun.py
@@ -61,7 +61,8 @@ class RBDDev(object):
         :return: status code and msg
         """
 
-        with rados.Rados(conffile=settings.config.cephconf) as cluster:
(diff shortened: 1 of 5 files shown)


ceph_iscsi_config/lun.py
```diff
@@ -867,6 +872,11 @@ def _add_dev_to_lio_user_rbd(self, in_wwn=None):
             if (settings.config.cephconf != '/etc/ceph/ceph.conf'):
                 cfgstring += ";conf={}".format(settings.config.cephconf)
 
+            if (settings.config.cluster_client_name != 'client.admin'):
+                dot_idx = settings.config.cluster_client_name.find('.')
```
dillaman: Nit: why not just use ```.split('.', 1)[1]```?

rjfd: yes, that's much better. pushed the fix

Issue #40, opened 2019-03-05
rpm: use systemd scriptlets macros


ktdreyer:
The RPM packaging for RHEL and Fedora currently hard-codes some systemd things. Like
```
BuildRequires:  systemd
```
or
```
/bin/systemctl --system daemon-reload &> /dev/null || :
/bin/systemctl --system enable rbd-target-gw &> /dev/null || :
/bin/systemctl --system enable rbd-target-api &> /dev/null || :
```

We can switch these out for scriptlets instead: https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_scriptlets

This will make it easier to get ceph-iscsi into Fedora if we follow the package guidelines here.

It will also make it easier to drop the dependency on systemd eventually when we run inside containers. [Fedora has a new %systemd_ordering macro for this](https://pagure.io/packaging-committee/issue/644), but I've not played around with it yet.

mikechristie:
I'll handle it this week. I am in the process of removing the need for that rbd-target-gw interaction right now. When that is removed, I will also fix up the reload/enable stuff above and test it all together.

ktdreyer:
No prob Mike, this is not urgent.

PR #51, opened 2019-03-17
Gateway management fixes and features

mikechristie:
This patchset adds several fixes and features related to gateway management:

1. Fixes several bugs dealing with gateway addition and startup.
2. Adds support to gwcli to remove gateways.
3. Fixes issue where rbd-target-api API call handling could race with rbd-target-gw startup/shutdown/restart by just having rbd-target-api do the initial target startup and shutdown instead of trying to play systemd dependency whack-a-mole. It does not fix races where metrics calls could race with API calls.

Jason, sorry that it is a lot of patches. A good chunk is just moving code around, and I dropped the cleanup and rename patches mentioned previously that were not directly needed for this set for another day :)

Note: The last patch requires some spec file magic, so we can disable and remove rbd-target-gw from old installs. It will also be related to this issue:

https://github.com/ceph/ceph-iscsi/issues/40


mikechristie:
One correction/update.

In the original push you might have seen there was a commit:

Subject: [PATCH 17/17] rm rbd-target-gw

Move metric requests to rbd-target-api.

where I tried to have rbd-target-api do 2 flask apps at the same time (metrics at port 9287 and REST API at 5000), but that is not supported by flask. I have done a repush and that is now dropped.

To fully kill off rbd-target-gw we would need to move the metrics code in rbd-target-gw into rbd-target-api under the one REST API flash app and users doing metrics calls need to then be ported to use port 5000. Do we want to do that and do we want to do that in this release?

dillaman:
> Do we want to do that and do we want to do that in this release?

It might be difficult to merge the two under a single flask app endpoint if that's what's required. The rationale being that the API port might be SSL secured, which implies that we would then need a non-SSL + SSL binding running concurrently (on two different ports).

PR #57, opened 2019-03-27
Adds `/api/clientinfo/<target_iqn>/<client_iqn>` endpoint (and more...)

ricardoasmarques:
This new endpoint will expose `alias`, `ip_address` and `state` of the client session.

Additionally, this PR also includes the following changes (related to the multi-target support):
- The `logged_in` information displayed in the `gwcli` will now display the information for the "current" target only (instead of displaying the info from the first session found regardless of the target that the client is connected in that session)
- The validation that checks if a client is logged in before allowing to delete him will now consider the target IQN.

### DIFF
--- a/ceph_iscsi_config/client.py
+++ b/ceph_iscsi_config/client.py
@@ -235,6 +235,34 @@ class GWClient(GWObject):
             self.change_count += 1
 
     @staticmethod
+    def get_client_info(target_iqn, client_iqn):
+        result = {
+            "alias": '',
+            "state": '',
+            "ip_address": []
+        }
+        iscsi_fabric = ISCSIFabricModule()
+        target = Target(iscsi_fabric, target_iqn, 'lookup')
+        for tpg in target.tpgs:
+            if tpg.enable:
+                for client in tpg.node_acls:
+                    if client.node_wwn != client_iqn:
+                        continue
+                    session = client.session
+                    if session is None:
+                        break
+                    result['alias'] = session.get('alias')
+                    state = session.get('state').upper()
+                    result['state'] = state
+                    ips = set()
+                    if state == 'LOGGED_IN':
+                        for conn in session.get('connections'):
+                            ips.add(conn.get('address'))
+                        result['ip_address'] = list(ips)
+                    break
+        return result
+
+    @staticmethod
     def define_clients(logger, config, target_iqn):
         """
         define the clients (nodeACLs) to the gateway definition
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -2111,6 +2111,73 @@ def _client(target_iqn, client_iqn):
             return jsonify(message="Client does not exist!"), 404
 
 
+@app.route('/api/clientinfo/<target_iqn>/<client_iqn>', methods=['GET'])
+@requires_restricted_auth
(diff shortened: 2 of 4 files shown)


ricardoasmarques:
I've pushed a new commit to this PR that changes the validation that checks if a client is logged in before allowing to delete him. This validation will now consider the target IQN.

In summary, if a client is connected to target "A", but **NOT** connected to target "B",
should **NOT** be possible to remove the client from target "A", but should
be possible to remove it from target "B".

ceph_iscsi_config/client.py
```diff
@@ -234,6 +234,30 @@ def define_client(self):
...
+            "alias": '',
+            "state": '',
+            "ip_address": []
+        }
+        iscsi_fabric = ISCSIFabricModule()
+        target = Target(iscsi_fabric, target_iqn, 'lookup')
+        for tpg in target.tpgs:
+            if tpg.enable:
+                for client in tpg.node_acls:
+                    if client.session and client.session['parent_nodeacl'].node_wwn == client_iqn:
```
mikechristie: I'm not sure I got the parent_nodeacl part. We know the session is through 'client' so we can just access the client.node_wwn. session['parent_nodeacl'] will always be the client we access the session through, so we can just do:

if client.node_wwn != client_iqn:
    continue;

# temp comment for review only. There can only be one client with client_iqn under each tpg, so once we find it, we know the search is done and we can break from the "for client in tpg.node_acls:" loop.

session = client.session:
if session is None:
    break

# copy session info to result
break

ceph_iscsi_config/client.py
```diff
@@ -234,6 +234,30 @@ def define_client(self):
...
+                        result['alias'] = client.session.get('alias')
+                        state = client.session.get('state').upper()
+                        result['state'] = state
+                        ips = set()
+                        if state == 'LOGGED_IN':
+                            for conn in client.session.get('connections'):
+                                ips.add(conn.get('address'))
+                            result['ip_address'] = list(ips)
+                            break
```
mikechristie: Was this break meant to break us out of the "for client in tpg.node_acls:" loop? Why is it only done if the state is logged in?

ricardoasmarques:
@mikechristie Addressed your comments, thanks for the hints.

rbd-target-api.py
```diff
@@ -2063,6 +2063,73 @@ def _client(target_iqn, client_iqn):
...
+    Examples:
+    curl --insecure --user admin:admin -X GET
+        http://192.168.122.69:5000/api/_clientinfo/
+        iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw/iqn.2003-01.com.redhat.iscsi-gw:iscsi-igw-client
+    """
+    if target_iqn not in config.config['targets']:
+        return jsonify(message="Target {} does not exist".format(target_iqn)), 400
+    target_config = config.config['targets'][target_iqn]
+    if client_iqn not in target_config['clients']:
+        return jsonify(message="Target {} does not exist".format(target_iqn)), 400
```
dillaman: Should this be "Client .... does not exist"?

ricardoasmarques: Fixed

PR #60, opened 2019-04-02
Adds '/api/gatewayinfo' endpoint

ricardoasmarques:
For now, this endpoint only returns the number of
active sessions on the local gateway.

### DIFF
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -1670,6 +1670,28 @@ def _targetinfo(target_iqn):
     }), 200
 
 
+@app.route('/api/gatewayinfo', methods=['GET'])
+@requires_restricted_auth
+def gatewayinfo():
+    """
+    Returns the number of active sessions on local gateway
+    **RESTRICTED**
+    Examples:
+    curl --insecure --user admin:admin -X GET
+        http://192.168.122.69:5000/api/gatewayinfo
+    """
+    local_gw = this_host()
+    if local_gw not in config.config['gateways']:
+        return jsonify(message="Gateway {} does not exist in configuration".format(local_gw)), 400
+    num_sessions = 0
+    for target_iqn, target in config.config['targets'].items():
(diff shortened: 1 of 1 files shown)


rbd-target-api.py
```diff
@@ -1670,6 +1670,43 @@ def _targetinfo(target_iqn):
...
+@requires_restricted_auth
+def gatewayinfo(gateway_name):
+    """
+    Returns the total number of active sessions on <gateway_name>
+    **RESTRICTED**
+    Examples:
+    curl --insecure --user admin:admin -X GET
+        http://192.168.122.69:5000/api/gatewayinfo/node1
+    """
+    resp_text, resp_code = call_api([gateway_name], '_gatewayinfo', '', http_method='get')
```
dillaman: Nit: should this validate that ```gateway_name``` is a valid gateway first? or perhaps just drop the ```gateway_name``` and force the caller to query the specific gateway API they are interested in and therefore avoid the indirection?

ricardoasmarques: I've removed the indirection.

rbd-target-api.py
```diff
@@ -1670,6 +1670,26 @@ def _targetinfo(target_iqn):
...
+    Returns the number of active sessions on local gateway
+    **RESTRICTED**
+    Examples:
+    curl --insecure --user admin:admin -X GET
+        http://192.168.122.69:5000/api/gatewayinfo
+    """
+    local_gw = this_host()
+    num_sessions = 0
+    for target_iqn, target in config.config['targets'].items():
+        if local_gw in target['portals']:
```
dillaman: Nit: should it return an error if the GW isn't in the portals list? 404 or 400?

ricardoasmarques: @dillaman I've added a validation that checks if `local_gw` is in `config['gateways']`

PR #63, opened 2019-04-04
Fix deletion of gateways with multiple IPs

ricardoasmarques:
PR https://github.com/ceph/ceph-iscsi/pull/51 introduced the support for deleting a gateway but doesn't consider the case where the gateway was created with multiple IPs.

This PRs will guarantee that when deleting a gateway, all IPs of that gateway are removed.

### How to reproduce:

Create a target:
`curl --insecure --user admin:admin -X PUT https://192.168.100.201:5001/api/target/iqn.2003-01.com.redhat.iscsi-gw0`

Create one gateway with a single IP:
`curl --insecure --user admin:admin -d ip_address=192.168.100.201 -dskipchecks=true -X PUT https://192.168.100.201:5001/api/gateway/iqn.2003-01.com.redhat.iscsi-gw0/node1`

Create another gateway with **two IPs** separated by a ',':
`curl --insecure --user admin:admin -d ip_address=192.168.100.202,192.168.121.120 -dskipchecks=true -X PUT https://192.168.100.201:5001/api/gateway/iqn.2003-01.com.redhat.iscsi-gw0/node2`

Delete the second gateway:
`curl --insecure --user admin:admin -X DELETE https://192.168.100.201:5001/api/gateway/iqn.2003-01.com.redhat.iscsi-gw0/node2`

One of the IPs from the second gateway will remain in the configuration:

```
{
  "created": "2019/04/04 17:32:35",
  "discovery_auth": {
    "mutual_password": "",
    "mutual_password_encryption_enabled": false,
    "mutual_username": "",
    "password": "",
    "password_encryption_enabled": false,
    "username": ""
  },
  "disks": {},
  "epoch": 4,
  "gateways": {
    "node1": {
      "active_luns": 0,
      "created": "2019/04/04 17:34:51",
      "updated": "2019/04/04 17:34:51"
    }
  },
  "targets": {
    "iqn.2003-01.com.redhat.iscsi-gw0": {
      "acl_enabled": true,
      "clients": {},
      "controls": {},
      "created": "2019/04/04 17:34:46",
      "disks": [],
      "groups": {},
      "ip_list": [
        "192.168.100.201",
        "192.168.100.202"
      ],
      "portals": {
        "node1": {
          "gateway_ip_list": [
            "192.168.100.201",
            "192.168.100.202"
          ],
          "inactive_portal_ips": [
            "192.168.100.202"
          ],
          "portal_ip_address": "192.168.100.201",
          "tpgs": 2
        }
      },
      "updated": "2019/04/04 17:35:05"
    }
  },
  "updated": "2019/04/04 17:35:06",
  "version": 8
}
```

And if we also delete the first gateway:

`curl --insecure --user admin:admin -X DELETE https://192.168.100.201:5001/api/gateway/iqn.2003-01.com.redhat.iscsi-gw0/node1`

The "lost" IP from the second gateway still remains in the configuration:

```
{
  "created": "2019/04/04 17:32:35",
  "discovery_auth": {
    "mutual_password": "",
    "mutual_password_encryption_enabled": false,
    "mutual_username": "",
    "password": "",
    "password_encryption_enabled": false,
    "username": ""
  },
  "disks": {},
  "epoch": 5,
  "gateways": {},
  "targets": {
    "iqn.2003-01.com.redhat.iscsi-gw0": {
      "acl_enabled": true,
      "clients": {},
      "controls": {},
      "created": "2019/04/04 17:34:46",
      "disks": [],
      "groups": {},
      "ip_list": [
        "192.168.100.202"
      ],
      "portals": {},
      "updated": "2019/04/04 17:35:29"
    }
  },
  "updated": "2019/04/04 17:35:29",
  "version": 8
}
```

### DIFF
--- a/ceph_iscsi_config/target.py
+++ b/ceph_iscsi_config/target.py
@@ -77,9 +77,9 @@ class GWTarget(GWObject):
                                   " any ip on this host")
                 return
 
-            self.active_portal_ip = list(matching_ip)[0]
+            self.active_portal_ips = list(matching_ip)
             self.logger.debug("active portal will use "
-                              "{}".format(self.active_portal_ip))
+                              "{}".format(self.active_portal_ips))
 
             self.gateway_ip_list = gateway_ip_list
             self.logger.debug("tpg's will be defined in this order"
@@ -88,7 +88,7 @@ class GWTarget(GWObject):
             # without gateway_ip_list passed in this is a 'init' or
             # 'clearconfig' request
             self.gateway_ip_list = []
-            self.active_portal_ip = []
+            self.active_portal_ips = []
 
         self.changes_made = False
         self.config_updated = False
@@ -177,25 +177,25 @@ class GWTarget(GWObject):
         :return: None
         """
 
+        index = 0
         for tpg in self.tpg_list:
             if tpg._get_enable():
                 for lun in tpg.luns:
                     try:
                         self.bind_alua_group_to_lun(config,
                                                     lun,
-                                                    tpg_ip_address=self.active_portal_ip)
+                                                    tpg_ip_address=self.active_portal_ips[index])
                     except CephiSCSIInval as err:
                         self.error = True
                         self.error_msg = err
                         return
 
                 try:
-                    NetworkPortal(tpg, normalize_ip_literal(self.active_portal_ip))
+                    NetworkPortal(tpg, normalize_ip_literal(self.active_portal_ips[index]))
                 except RTSLibError as e:
                     self.error = True
                     self.error_msg = e
-                else:
-                    break
+                index += 1
 
     def clear_config(self, config):
         """
@@ -249,7 +249,7 @@ class GWTarget(GWObject):
 
             self.logger.debug("(Gateway.create_tpg) Added tpg for portal "
                               "ip {}".format(ip))
-            if ip == self.active_portal_ip:
+            if ip in self.active_portal_ips:
                 if self.enable_portal:
                     NetworkPortal(tpg, normalize_ip_literal(ip))
                 tpg.enable = True
@@ -383,7 +383,7 @@ class GWTarget(GWObject):
         # they do not have a common gw the owning gw may not exist here.
         # The LUN will just have all ANO paths then.
         if gw_config:
-            if gw_config["portal_ip_address"] == tpg_ip_address:
+            if gw_config["portal_ip_addresses"][0] == tpg_ip_address:
                 is_owner = True
 
         try:
@@ -575,7 +575,8 @@ class GWTarget(GWObject):
                         continue
 
                     inactive_portal_ip = list(self.gateway_ip_list)
-                    inactive_portal_ip.remove(remote_gw_config["portal_ip_address"])
+                    for portal_ip_address in remote_gw_config["portal_ip_addresses"]:
+                        inactive_portal_ip.remove(portal_ip_address)
                     remote_gw_config['gateway_ip_list'] = self.gateway_ip_list
                     remote_gw_config['tpgs'] = len(self.tpg_list)
                     remote_gw_config['inactive_portal_ips'] = inactive_portal_ip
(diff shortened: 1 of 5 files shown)


ricardoasmarques:
@dillaman `portal_ip_address` was changed to a `portal_ip_addresses` list.

mikechristie:
I didn't know multiple IPs was supported. What commit was it added in? Did it come in with the multiple target support?

I think some other stuff is broken or might just be sort-of working by accident.

mikechristie:
@ricardoasmarques One other question. Are you only using multiple IPs with the SUSE kernel target_core_rbd module? You do not use it with the target_core_user and tcmu-runner do you?

ceph_iscsi_config/target.py
```diff
@@ -177,25 +177,25 @@ def enable_active_tpg(self, config):
         :return: None
         """
 
+        index = 0
```
mikechristie: I think code wise the patch looks ok.

ceph-iscsi and tcmu-runner though assumed that there would only be one iscsi portal per target port group. If we hit this loop more than once then I think bind_alua_group_to_lun is going to fail because we pass in the same tpg tag to ALUATargetPortGroup and it will try to do a create for the same tpg twice.

I think failing might be ok though, because the user would get an error while trying to set it up and so we would not allow it.

@ricardoasmarques you guys do not use target_core_user/tcmu-runner in active/passive mode right? So it should be ok for both of our types of set ups I think?

ricardoasmarques:
> I didn't know multiple IPs was supported. What commit was it added in? Did it come in with the multiple target support?
> 
> I think some other stuff is broken or might just be sort-of working by accident.

This was already available before the support for multiple targets, but it's curious that you cannot do this using `gwcli`, you have to call the API directly so maybe this "feature" was not intentionally implemented.

Shouldn't we change the `gwcli` gateway creation command to allow the user to specify multiple IPs, for consistency with the API?

ricardoasmarques:
> @ricardoasmarques One other question. Are you only using multiple IPs with the SUSE kernel target_core_rbd module? You do not use it with the target_core_user and tcmu-runner do you?

**Ceph Dashboard** allows the user to configure multiple IPs regardless of the selected `backstore`, so both `kernel` and `tcmu-runner` are now able to configure it.

ceph_iscsi_config/target.py
ricardoasmarques: @mikechristie Using multiple IPs in `lrbd` will create multiple portals in the same `tpg`.

For instance, in `ceph-iscsi` you will have:

```
tpg2
  - portals
    192.168.100.202:3160
tpg3
  - portals
    192.168.121.123:3160
```

and in lrbd you will have:

```
tpg2
  - portals
    192.168.100.202:3160
    192.168.121.123:3160
```

Isn't this equivalent? Should we change `ceph-iscsi` to create multiple portals in the same tpg?

mikechristie: I think multiple portals in the same tpg will work for what we support in RHCS. I am doing some testing now.

mikechristie:
> > I didn't know multiple IPs was supported. What commit was it added in? Did it come in with the multiple target support?
> > I think some other stuff is broken or might just be sort-of working by accident.
> 
> This was already available before the support for multiple targets, but it's curious that you cannot do this using `gwcli`, you have to call the API directly so maybe this "feature" was not intentionally implemented.
> 

It was not intentional for what we currently support so that is why we never added gwcli support. It was probably something we did early on when we were doing single node non target_core_user based support originally. We then later dropped that, but we never cleaned up code like this.

The problem is that some SCSI commands require the port info, but target_core_user and tcmu-runner do not have it since they are too far up the stack. We were limiting it to one enabled tpg with one portal per node for target_core_user/runner setups, because we then know the command could only come from the enabled one.

> Shouldn't we change the `gwcli` gateway creation command to allow the user to specify multiple IPs, for consistency with the API?

It might be a little more difficult. Other changes might be needed to fully support it with target_core_user.

For multiple enabled tpgs support we need kernel and tcmu-runner changes.

Multiple portals per tpg might work, because we do not support and initiators have not required support for any type of command that would require us to return or use info for a specific portal in a tpg. For example, the relative port id would be the same for all portals in that tpg so we would be ok for that command. There are commands that might specify/require the ip address of a portal, but we do not support them. I am still checking the specs. And, there also might be a bug with some versions of windows where if you return the same relative port id for a LUN accessed through different iscsi sessions, then windows will crash.

ricardoasmarques:
@dillaman Any other changes that should be included in this PR?

dillaman:
@ricardoasmarques Sorry, I've been distracted by other things. @mikechristie, assuming you think it's good-to-go, feel free to merge.

mikechristie:
This was ok with me. We will fix up the general multiple ip support in another PR.

PR #66, opened 2019-04-16
Fix 'targetinfo' endpoint when iSCSI port is down

ricardoasmarques:
If iSCSI port is down or not listening, the 'call_api' will return an HTTP 503 code.
In the context of 'targetinfo' endpoint this should not be considered an error because if this happens we can assume that there are no active sessions.

### DIFF
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -1638,6 +1638,10 @@ def targetinfo(target_iqn):
     gateways = target_config['portals']
     num_sessions = 0
     for gateway in gateways.keys():
+        target_state = target_ready([gateway])
+        if target_state.get('status_api') == 'UP' and target_state.get('status_iscsi') == 'DOWN':
+            # If API is 'up' and iSCSI is 'down', there are no active sessions to count
+            continue
         resp_text, resp_code = call_api([gateway], '_targetinfo', target_iqn, http_method='get')
         if resp_code != 200:
             return jsonify(message="{}".format(resp_text)), resp_code
@@ -2489,6 +2493,8 @@ def target_ready(gateway_list):
     """
     http_mode = 'https' if settings.config.api_secure else 'http'
     target_state = {"status": 'OK',
+                    "status_iscsi": 'UP',
+                    "status_api": 'UP',
                     "summary": ''}
 
     for gw in gateway_list:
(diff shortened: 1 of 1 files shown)


mikechristie:
Oh no, wait I might take that ACK back.

gateways is the 'portals' list in the target_config, and portals is not the iscsi RFC portal but the IP address the gateway node listens for REST requests on right? If you have multiple iscsi RFC portals then one could be the same IP ass the REST IP, but you would have other iscsi RFC portals that have different IPs. If the IPs are on different networks/nics/hardware then the _targetinfo->call_api->ping could fail to the REST IP but some iscsi RFC portals might be up and still have iscsi sessions. In this case we would return the wrong number of sessions with the patch right? If that is correct, are we just saying we don't care in that case that the session count could be wrong?

ricardoasmarques:
> ...
> ` for gateway in gateways.keys():`
> 
> gateways is the 'portals' list in the target_config, and portals is not the iscsi RFC portal but the IP address the gateway node listens for REST requests on right?
> ...

@mikechristie `gateway` var is the hostname, not the IP so I think the scenario you described should not happen. Or am I missing something?

mikechristie:
> @mikechristie `gateway` var is the hostname, not the IP so I think the scenario you described should not happen. Or am I missing something?

Maybe we are talking about different issues.

My issue is not related to it being a hostname or IP. The hostname gets resolved to a IP eventually when we do:

targetinfo call_api(gateway...) -> target_ready -> normalize_ip_literal(gateway) -> REST _ping call

My issue was that a ping above might fail, but sessions could be up via different iSCSI RFC portals on that node.

ricardoasmarques:
@mikechristie but in that case, the ping will fail with an error different than `503`, so the error will be propagated to the REST API client, right?

The only change included in this patch is the case when the API is available, but the [iscsi_active](https://github.com/ceph/ceph-iscsi/blob/master/rbd-target-api.py#L2446) returns `False`, which will result in a `503` HTTP code.

In this patch, I 'm assuming that even when using multiple RFC portals, [iscsi_active](https://github.com/ceph/ceph-iscsi/blob/master/rbd-target-api.py#L2446) will not return `False` if there are active sessions. Is this assumption wrong? If this assumption is wrong maybe we have to bypass the `call_api` method and do the REST request directly from the `targetinfo` method? Any other idea?

mikechristie:
> @mikechristie but in that case, the ping will fail with an error different than `503`, so the error will be propagated to the REST API client, right?

targetinfo will see 503 for every possible error.

If iscsi_active does not see a iscsi target up and listening then it will return 503. We agree on that.

I am also saying, it looks like anytime target_ready() does not return 'OK' for the status, call_api returns 503 too. So for cases like where the _ping request does not even make it to the other host's rbd-target-api daemon because rbd-target-api is not up, the other host is not reachable on the IP we are trying to ping, whatever error there is, target_ready will set status to "NOTOK". call_api then returns 503 for all errors here:

call_api():

```
    target_state = target_ready(gateway_list)
    if target_state.get('status') != 'OK':
        return ('failed, gateway(s) unavailable:'
                '{}'.format(target_state.get('summary'))), 503
```

ricardoasmarques:
@mikechristie I finally understood what you mean.

I've now updated this PR with a different approach where we explicitly check the API and iSCSI status.

Please review again, thanks.

rbd-target-api.py
```diff
@@ -2500,15 +2504,21 @@ def target_ready(gateway_list):
             api.get()
```
mikechristie: When we initialize target_state at the beginning of this function, do we need to also init status_iscsi and status_api to both 'UP'?

ricardoasmarques: Fixed, thanks.

mikechristie:
Looks ok to me.

PR #78, opened 2019-05-06
Update spec to use systemd scriplets

mikechristie:
This updates our spec to use systemd scriptlets for fedora/RHEL.
They are described here:

https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/

There is one major behavior change with this patch. I followed the doc
assuming we wanted to get into fedora/RHEL. The problem is that this doc:

https://docs.fedoraproject.org/en-US/packaging-guidelines/DefaultServices/

lists requirements for enabling the service when the package is
installed and we do not meet them, so this patch does not enable us
by default.

However, I can just update the docs for this. The upstream/RHCS docs
already state we must enable rbd-target-api so we just need to add
a line about also enabling rbd-target-gw.

### DIFF
--- a/ceph-iscsi.spec
+++ b/ceph-iscsi.spec
@@ -68,7 +68,8 @@ Requires:       python3-configshell
 %endif
 %endif
 
-BuildRequires:  systemd
+BuildRequires: systemd-rpm-macros
+%{?systemd_requires}
 
 %description
 Python package providing the modules used to handle the configuration of an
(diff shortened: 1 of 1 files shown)


ceph-iscsi.spec
```diff
@@ -68,7 +68,8 @@ Requires:       python3-configshell
 %endif
 %endif
 
-BuildRequires:  systemd
+BuildRequires: systemd-rpm-macros
+%{?systemd_requires}
```
mikechristie: @ricardoasmarques 

Will this be ok for SUSE? I saw you have similar scriplets here

https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines

but in "(Build) Requirements" it looks like you also have to check for a specific suse version. Is that check needed or ceph-iscsi only in the newer version?

mikechristie:
This fixes issue

https://github.com/ceph/ceph-iscsi/issues/40

ceph-iscsi.spec
ricardoasmarques: @smithfarm can you help here? thanks.

smithfarm: @mikechristie The non-systemd versions of openSUSE are all EOL now, so that document is outdated. In other words, this line is OK for SUSE :-)

ceph-iscsi.spec
```diff
@@ -68,7 +68,8 @@ Requires:       python3-configshell
 %endif
 %endif
 
-BuildRequires:  systemd
+BuildRequires: systemd-rpm-macros
```
smithfarm: this line is also OK for SUSE

ceph-iscsi.spec
```diff
@@ -132,22 +133,26 @@ ln -s service %{buildroot}%{_sbindir}/rcrbd-target-api
...
+%systemd_post rbd-target-api.service
 %endif
 %if 0%{?suse_version}
 %service_add_post rbd-target-gw.service rbd-target-api.service
 %endif
 
 %preun
+%if 0%{?fedora} || 0%{?rhel}
+%systemd_prerun rbd-target-gw.service
+%systemd_prerun rbd-target-api.service
```
smithfarm: Should be `%systemd_preun` (not "prerun") here I think

mikechristie: You are right. Thanks. Fixed and repushed.

PR #79, opened 2019-05-06
Journaling is a supported RBD feature

ricardoasmarques:
Note that, despite being supported, 'Journaling' is not enabled by default.

### DIFF
--- a/ceph_iscsi_config/lun.py
+++ b/ceph_iscsi_config/lun.py
@@ -26,17 +26,8 @@ __author__ = 'pcuzner@redhat.com'
 
 class RBDDev(object):
 
-    supported_features_list = {
-        USER_RBD: [
-            'RBD_FEATURE_LAYERING',
-            'RBD_FEATURE_EXCLUSIVE_LOCK',
-            'RBD_FEATURE_OBJECT_MAP',
-            'RBD_FEATURE_FAST_DIFF',
-            'RBD_FEATURE_DEEP_FLATTEN',
-            'RBD_FEATURE_JOURNALING',
-            'RBD_FEATURE_DATA_POOL',
-            'RBD_FEATURE_OPERATIONS'
-        ]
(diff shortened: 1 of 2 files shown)


ceph_iscsi_config/lun.py
```diff
@@ -26,7 +26,18 @@
 
 class RBDDev(object):
 
-    rbd_feature_list = {
+    supported_features_list = {
```
dillaman: Might as well add ```RBD_FEATURE_DATA_POOL``` and ```RBD_FEATURE_OPERATIONS``` while in here. Or if you don't need to exclude certain features for the SUSE implementation, perhaps just remove it since really the only thing the default implementation cares about it exclusive-lock.

ricardoasmarques: @dillaman I've just added `RBD_FEATURE_DATA_POOL` and `RBD_FEATURE_OPERATIONS` to the supported features list (because SUSE implementation has to exclude, for instance, `RBD_FEATURE_JOURNALING`).

One additional question, shouldn't we also add the `RBD_FEATURE_STRIPINGV2` to the supported features list?

dillaman: Ah -- indeed. Perhaps it's just easier to list the excluded features so that we don't need to keep changing this. SUSE would have OBJECT_MAP/FAST_DIFF/JOURNALING/MIGRATION

ricardoasmarques: Agree, a list of excluded features is easier to maintain, I'll change it (and do the corresponding changes on ceph dashboard, because ceph dashboard is now filtering the RBD images based on the supported features returned by `/api/settings` endpoint).

ricardoasmarques: @dillaman I'm now using a list of unsupported features, instead of a list of supported features. Please review again, thanks.

PR #84, opened 2019-05-13
Add support for enabling FQDN

ricardoasmarques:
With this PR will be possible to use FQDNs instead of host shortnames.

If `fqdn_enabled` setting is `true` (defaults to `false`), then FQDN will be used, otherwise, nothing changes from the previous implementation.

### DIFF
--- a/ceph_iscsi_config/common.py
+++ b/ceph_iscsi_config/common.py
@@ -1,4 +1,5 @@
 import rados
+import socket
 import time
 import json
 import traceback
@@ -55,7 +56,7 @@ class Config(object):
                                       'mutual_username': '',
                                       'mutual_password': '',
                                       'mutual_password_encryption_enabled': False},
-                   "version": 9,
+                   "version": 10,
                    "epoch": 0,
                    "created": '',
                    "updated": ''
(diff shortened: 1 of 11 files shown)


ricardoasmarques:
FYI, the main reason behind this PR is to support SSL certificates that use domain names.

LenzGr:
LGTM, thanks!

gwcli/utils.py
```diff
@@ -36,7 +36,7 @@ def this_host():
     return the local machine's fqdn or shortname
     """
     fqdn_enabled = settings.config.fqdn_enabled
-    return socket.gethostname() if fqdn_enabled else socket.gethostname().split('.')[0]
+    return socket.getfqdn() if fqdn_enabled else socket.gethostname().split('.')[0]
```
mikechristie: Can we just kill this one and convert gwcli/rbd-tagret-api to use the ceph_iscsi_config/utils one?

I think it was just an oversight we had 2 functions.

ricardoasmarques: @mikechristie I've drop gwcli.utils::this_host in favor of ceph_iscsi_config.utils::this_host

mikechristie:
Hey @ricardoasmarques 

I wanted to make a new release. Did you want this feature in it? And if so, did you need help fixing up the patch? I can do it for you, if you are busy.

ricardoasmarques:
@dillaman @mikechristie sorry for the late reply, I was on vacation last week.

Storing if FQDN is enabled or not in the config will require to read the config every time we need to call `this_host` function, which will require to read the config in places where we are not reading the config ATM (e.g. in the `rbd-target-gw` service).

To avoid this dependency, I would like to propose a different approach where we upgrade the config to replace gateways short names by the corresponding FQDNs and do not support to disable it. I mean, `ceph-iscsi` will always use FQDNs instead of short names.

I'll submit my proposal today.

ricardoasmarques:
@mikechristie @dillaman please review again.

In this new approach, we are completely migrating from host shot names to FQDNs, so we don't have to maintain both. Do you agree?

ricardoasmarques:
9.5 is a temporary version, dashboard will not work until config is fully upgraded to version 10

ceph_iscsi_config/common.py
```diff
@@ -315,6 +316,44 @@ def _upgrade_config(self):
...
+                    if disk.get('allocating_host') == this_shortname:
+                        disk['allocating_host'] = this_fqdn
+                    if disk.get('owner') == this_shortname:
+                        disk['owner'] = this_fqdn
+                    self.update_item("disks", disk_id, disk)
+                gateways_upgraded.append(this_fqdn)
+                self.update_item("gateways_upgraded", None, gateways_upgraded)
+            if any(gateway_name not in gateways_upgraded
+                   for gateway_name in self.config['gateways'].keys()):
+                # upgrade vom v9 to v10 is still in progress, some gateways are not upgraded yet
```
mikechristie: Just a trivial comment. "vom" should be "from".

ricardoasmarques:
@mikechristie Typo is now fixed.

ricardoasmarques:
See https://github.com/ceph/ceph/pull/28720 for a list of steps that dashboard users should perform when upgrading to this `ceph-iscsi` version.

Issue #95, opened 2019-05-29
ACL settings are not applied to all TPGs when using multiple IPs


ricardoasmarques:
Using the `ceph-iscsi` REST API (through Ceph Dashboard) I've created the following configuration (note that `node1` has 2 IPs):

![Screenshot from 2019-05-29 12-23-48](https://user-images.githubusercontent.com/14297426/58553749-a3743b00-820d-11e9-8b71-4b10903b0295.png)

But looking into the `targetcli` output, I see that ACL settings were not applied to TPG2:

![Screenshot from 2019-05-29 12-24-01](https://user-images.githubusercontent.com/14297426/58553835-dae2e780-820d-11e9-8ff5-3bdd6d574c03.png)

PR #97, opened 2019-05-30
Adds support for multiple IPs per gateway on REST API

ricardoasmarques:
With this PR, `ceph-iscsi` REST API will correctly handle multiple IPs on the same host by creating one portal for each IP:

Screenshot from 2019-05-30 10-50-04

For more information about the behaviour before this PR, see the corresponding issue (https://github.com/ceph/ceph-iscsi/issues/95).

Fixes: https://github.com/ceph/ceph-iscsi/issues/95

### DIFF
--- a/ceph_iscsi_config/target.py
+++ b/ceph_iscsi_config/target.py
@@ -39,7 +39,7 @@ class GWTarget(GWObject):
     # gwcli to get/set all tpgs/clients under the target instead of per obj.
     SETTINGS = TPG_SETTINGS + GWClient.SETTINGS
 
-    def __init__(self, logger, iqn, gateway_ip_list, enable_portal=True):
+    def __init__(self, logger, config, iqn, gateway_ip_list, enable_portal=True):
         """
         Instantiate the class
         :param iqn: iscsi iqn name for the gateway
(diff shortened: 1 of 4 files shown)


mikechristie:
Sorry @ricardoasmarques, I am behind at work. I should be done testing these PRs and fixing up tcmu-runner to work with them by friday or more likely sunday.

ricardoasmarques:
@mikechristie thanks for looking into this, and thanks for the update.

ceph_iscsi_config/target.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, logger, iqn, gateway_ip_list, enable_portal=True):
 
         self.enable_portal = enable_portal  # boolean to trigger portal IP creation
         self.logger = logger                # logger object
+        self.config = config
```
mikechristie: The classes that are based on the GWObject have a Config object already.

mikechristie:
Just adding a note. It turns out we do not need any tcmu-runner changes even for the multiple portal per tpg implementation in this PR. tcmu-runner does not support just the right amount of SCSI features :)

ceph_iscsi_config/target.py
ricardoasmarques: Good spotting. This is fixed now, thanks.

PR #98, opened 2019-05-30
Adds support for multiple IPs per gateway on gwcli

ricardoasmarques:
When using `ceph-iscsi` REST API directly (e.g. from Ceph Dashboard) multiple IPs for a single gateway is already supported (if `skipchecks` is set to `true`), but `gwcli` only accepts one IP.

This PR changes the `gwcli` to also accept multiple comma separated IPs (without the need to set `skipchecks=true`).

E.g.:
```
/iscsi-target...0343/gateways> create node1 192.168.100.201,192.168.121.236 
```

Note that `configshell-fb` only supports `,` since https://github.com/open-iscsi/configshell-fb/commit/341018797fcb47bdeda1a5c87266d4b14f51c945 which means that multiple IPs on `gwcli` requires `configshell-fb` >= `v1.1.fb25` (single IP gateways will still work when using older versions of `configshell-fb`)

### DIFF
--- a/ceph-iscsi.spec
+++ b/ceph-iscsi.spec
@@ -48,7 +48,7 @@ Requires:       python-rtslib >= 2.1.fb68
 Requires:       rpm-python >= 4.11
 Requires:       python-cryptography
 Requires:       python-flask >= 0.10.1
-Requires:       python-configshell
+Requires:       python-configshell >= 1.1.fb25
 %if 0%{?rhel} == 7
 Requires:       pyOpenSSL
 %else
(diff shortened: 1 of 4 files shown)


dillaman:
@ricardoasmarques Perhaps the ```ceph-iscsi.spec``` should be tweaked for the new configshell restriction?

ricardoasmarques:
@dillaman I've added the `configshell` version restriction to the `ceph-iscsi.spec` file.

dillaman:
@ricardoasmarques can you please rebase?

ricardoasmarques:
> @ricardoasmarques can you please rebase?

rebased

PR #104, opened 2019-06-10
Remove config hash check validation

ricardoasmarques:
While trying to use different `cluster_client_name` for each gateway I've noticed that, when adding a new gateway, `ceph-iscsi` is forcing the new gateway to have the same `/etc/ceph/iscsi-gateway.cfg` file hash (same "comments", same `cluster_client_name`, same `logger_level`, etc.).

I don't understand why we have this restriction, so this PR is removing this validation.

If we really need this check, maybe we should find a different way to implement it in order to allow some settings to be different across gateways. Thoughts?

### DIFF
--- a/ceph_iscsi_config/utils.py
+++ b/ceph_iscsi_config/utils.py
@@ -5,7 +5,6 @@ import rados
 import rbd
 import re
 import datetime
-import hashlib
 import os
 import rpm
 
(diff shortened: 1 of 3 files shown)


dillaman:
There are a lot of settings that would cause issues if they were out-of-sync between gateways. Perhaps instead of hashing the file, it could json-ify the important in-memory settings and compare hashes (e.g. min gateways, secure, ports, LIO overrides)

ricardoasmarques:
Closing in favor of https://github.com/ceph/ceph-iscsi/pull/109

PR #105, opened 2019-06-10
spec: fix rhel conditional blocks

smithfarm:
35f0532990356e00e79656a8cdae379885be8725 introduced the following conditional:

```    
%if 0%{?rhel} < 8
...
%endif
```

This is problematic because the conditional will evaluate to true if the macro
`%rhel` is undefined, which is the case on openSUSE for example.

### DIFF
--- a/ceph-iscsi.spec
+++ b/ceph-iscsi.spec
@@ -68,7 +68,7 @@ Requires:       python3-configshell
 %endif
 %endif
 
-%if 0%{?rhel} < 8
+%if 0%{?rhel} == 7
 BuildRequires: systemd
 %else
 BuildRequires: systemd-rpm-macros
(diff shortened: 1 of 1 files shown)


ceph-iscsi.spec
```diff
@@ -68,7 +68,7 @@ Requires:       python3-configshell
 %endif
 %endif
 
-%if 0%{?rhel} < 8
+%if 0%{?rhel} && 0%{?rhel} < 8
```
dillaman: Maybe just change these to ```0%{?rhel} == 7``` to avoid the need to wrap the extra conditionals?

smithfarm: If you're sure ceph-iscsi will never be built for 6 and below, then :+1:

dillaman: 100% sure -- not supported.

smithfarm:
@dillaman Please take another look?

PR #107, opened 2019-06-10
SUSE is now supported

ricardoasmarques:
This PR adds SUSE to the list of supported OS.

By default, we are now relying on `/etc/os-release` to get the release version, but for 'redhat', we are still using `platform.linux_distribution` because it has more precision (it also returns the release minor version), but note that `platform` is deprecated so this should be adapted in the future. 

Additionally, the `rpm` version checks on the REST API were moved to the `spec` file only.

### DIFF
--- a/ceph-iscsi.spec
+++ b/ceph-iscsi.spec
@@ -37,6 +37,7 @@ BuildArch:      noarch
 Obsoletes:      ceph-iscsi-config
 Obsoletes:      ceph-iscsi-cli
 
+Requires:       tcmu-runner >= 1.3.0
 %if 0%{?with_python2}
 BuildRequires:  python2-devel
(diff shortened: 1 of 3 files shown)


ceph-iscsi.spec
```diff
@@ -37,6 +37,7 @@ BuildArch:      noarch
 Obsoletes:      ceph-iscsi-config
 Obsoletes:      ceph-iscsi-cli
 
+Requires:       tcmu-runner >= 1.3.0
```
dillaman: Nit: probably ```>= 1.4.0``` now

ricardoasmarques: Fixed. I've also created a PR to update the ceph documentation accordingly - https://github.com/ceph/ceph/pull/28494

PR #109, opened 2019-06-10
Fix config hash check validation 

ricardoasmarques:
Only some settings are required to be in sync between gateways so, instead of forcing the whole config file to be the same in all gateways, this PR will only consider the settings that really have to be in sync.

With this PR will be possible, for instance, to have different `cluster_client_name` on each gateway.

### DIFF
--- a/ceph_iscsi_config/settings.py
+++ b/ceph_iscsi_config/settings.py
@@ -120,13 +120,9 @@ class Settings(object):
                 "logger_level": logging.DEBUG
                 }
 
-    sync_required = ["cluster_name",
-                     "pool",
-                     "api_port",
-                     "api_secure",
-                     "minimum_gateways",
-                     "prometheus_port"
-                     ]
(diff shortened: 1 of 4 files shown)


ceph_iscsi_config/settings.py
```diff
@@ -118,6 +120,14 @@ def normalize(k, v):
                 "logger_level": logging.DEBUG
                 }
 
+    sync_required = ["cluster_name",
```
dillaman: Perhaps this should instead be a blacklist for excluding specific configuration options that don't matter if they differ between gateways. Right now I think this list is missing lots of important things that shouldn't be allowed to diverge: i.e. api_ssl_verify, api_user, api_password, .....

ricardoasmarques: Fixed. I'm now using a blacklist of setting that should be excluded from the "hash" calculation.

PR #110, opened 2019-06-10
Fix systemd rbd-target-gw deps

mikechristie:
Since we moved startup to rbd-target-api, it no longer has a dep on
rbd-target-gw, so drop the systemd dependencies.conf.

This fixes cycle detected bugs that depending on the systemd version
prevent rbd-target-api or rbd-target-gw from starting.

### DIFF
--- a/ceph-iscsi.spec
+++ b/ceph-iscsi.spec
@@ -129,7 +129,6 @@ mkdir -p %{buildroot}%{_mandir}/man8
 install -m 0644 gwcli.8 %{buildroot}%{_mandir}/man8/
 gzip %{buildroot}%{_mandir}/man8/gwcli.8
 mkdir -p %{buildroot}%{_unitdir}/rbd-target-gw.service.d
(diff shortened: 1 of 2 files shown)


dillaman:
@mikechristie Does the ```rbd-target-api``` need a dependency file for ```rbd-target-gw``` now since the role has been reversed?

mikechristie:
Currently, if rbd-target-api is down but -gw is up then we get empty stats. If we rebind them (revert https://github.com/ceph/ceph-iscsi/commit/03f8ca9a2130376bb7bae62d50a31fedb4228209#diff-38d8c1aef72556027f16811f8cb051e5 since the problem was really the dependencies.conf file) so -gw goes down when -api does, then the caller will get an error indicating the service is down which I think might be easier for callers to understand than getting a empty return value.

If we do rebind them though, do we need the dependency file or can we just add everything to only the systemd unit file? It was not clear to me why we needed to do the dependencies.conf file to override our own systemd unit file settings.

PR #112, opened 2019-06-12
add support for LUN qos

wuxingyi:
qos is important for LUNs, in qemu-librbd scene we can use qemu to
throttle iops/bps in server side, so the vm user can't use to much
iops/bps than we provide him.

we can't control rbd image qos before ceph nautilus introduced this
feature, tcmu-runner running librbd will help us control block storage rate.

Signed-off-by: wuxingyi <wuxingyi2015@outlook.com>

### DIFF
--- a/ceph_iscsi_config/lun.py
+++ b/ceph_iscsi_config/lun.py
@@ -158,6 +158,72 @@ class RBDDev(object):
                         else:
                             self.changed = True
 
+    def rbd_setqos(self, qos):
+        """
+        set qos parameters
+        :param qos: qos parameters (dict) qos parameters to control the iops/bps rate
+        """
+        with rados.Rados(conffile=settings.config.cephconf,
+                         name=settings.config.cluster_client_name) as cluster:
+            with cluster.open_ioctx(self.pool) as ioctx:
+                with rbd.Image(ioctx, self.image) as rbd_image:
+                    lock_info = rbd_image.list_lockers()
+                    if lock_info:
+                        lockers = lock_info.get("lockers")
+                        for holder in lockers:
+                            try:
+                                rbd_image.break_lock(holder[0], holder[1])
+                            except Exception:
+                                self.error = True
+                                self.error_msg = ("failed to break lock for {}".format(self.image))
+                                return
+
+                        try:
+                            rbd_image.lock_acquire(rbd.RBD_LOCK_MODE_EXCLUSIVE)
+                        except Exception:
+                            self.error = True
+                            self.error_msg = ("failed to acquire lock for {}".format(self.image))
+                            return
+
+                        for key, limit in qos.items():
+                            try:
+                                rbd_image.metadata_set(key, limit)
+                            except Exception:
+                                self.error = True
+                                self.error_msg = ("metadata_set failed for param {}".format(key))
+                                rbd_image.lock_release()
+                                return
+                    else:
+                        for key, limit in qos.items():
+                            try:
+                                rbd_image.metadata_set(key, limit)
+                            except Exception:
+                                self.error = True
+                                self.error_msg = ("metadata_set failed for param {}".format(key))
+                                return
+        return
+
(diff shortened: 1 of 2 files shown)


ceph_iscsi_config/lun.py
```diff
@@ -158,6 +158,72 @@ def rbd_size(self):
...
+        with rados.Rados(conffile=settings.config.cephconf,
+                         name=settings.config.cluster_client_name) as cluster:
+            with cluster.open_ioctx(self.pool) as ioctx:
+                with rbd.Image(ioctx, self.image) as rbd_image:
+                    lock_info = rbd_image.list_lockers()
+                    if lock_info:
+                        lockers = lock_info.get("lockers")
+                        for holder in lockers:
+                            try:
+                                rbd_image.break_lock(holder[0], holder[1])
```
mikechristie: Does break_lock also blacklist the current lock holder or is that lock_break()?

If it does not blacklist the holder then we are going to end up in a situation where a client thinks it has the lock and is doing IO but is never properly fenced off.

The break_lock use case in lun.py is only when a lun is first added and it only breaks it's own locks and not other nodes, so it does not worry about blacklisting.

ceph_iscsi_config/lun.py
```diff
@@ -158,6 +158,72 @@ def rbd_size(self):
...
+                            return
+
+                        for key, limit in qos.items():
+                            try:
+                                rbd_image.metadata_set(key, limit)
+                            except Exception:
+                                self.error = True
+                                self.error_msg = ("metadata_set failed for param {}".format(key))
+                                rbd_image.lock_release()
+                                return
```
mikechristie: Don't you want to release the lock here too?

ceph_iscsi_config/lun.py
```diff
@@ -158,6 +158,72 @@ def rbd_size(self):
...
+                    else:
+                        for key, limit in qos.items():
+                            try:
+                                rbd_image.metadata_set(key, limit)
+                            except Exception:
+                                self.error = True
+                                self.error_msg = ("metadata_set failed for param {}".format(key))
+                                return
```
mikechristie: Do you need the lock at all, or why do you only take it if there are other holders but not in this case?

ceph_iscsi_config/lun.py
```diff
@@ -158,6 +158,72 @@ def rbd_size(self):
...
+        return
```
mikechristie: No return needed.

ceph_iscsi_config/lun.py
```diff
@@ -158,6 +158,72 @@ def rbd_size(self):
...
+                        ml = rbd_image.metadata_list()
+                        qos = {}
+                        for key, limit in ml:
+                            if key.startswith('conf_rbd_qos_'):
+                                qos[key] = limit
+                        return qos
+                    except Exception as err:
+                        self.error = True
+                        self.error_msg = ("get qos param failed, error is {}".format(err))
+        return
```
mikechristie: No return is needed.

rbd-target-api.py
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-
```
mikechristie: Drop this from the patch. It is not related. And, I do not know if it is the proper coding style, but if it is you could fix it in the other code too in a separate patch.

rbd-target-api.py
```diff
@@ -931,6 +930,84 @@ def get_disks():
...
+                except ValueError:
+                    return jsonify(message="{} must be integer".format(image_id, limit)), 400
+                if int(v) < 0:
+                    return jsonify(message="{} must be positive".format(image_id, limit)), 400
+
+                qos[limit] = v
+
+        if qos:
+            try:
+                rbd_image = RBDDev(image, 0, LUN.DEFAULT_BACKSTORE, pool)
```
mikechristie: You can get the backstore from the config object here and below.

PR #115, opened 2019-06-27
Fix spec for RHEL 8 builds

mikechristie:
systemd-rpm-macros was recently broken off from systemd into its own
package. It is in Fedora but not yet in RHEL 8. This checks for RHELs
less than 8 assuming that in RHEL 9 we will have syncd up with fedora
and will have systemd-rpm-macros.

### DIFF
--- a/ceph-iscsi.spec
+++ b/ceph-iscsi.spec
@@ -75,7 +75,7 @@ Requires:       python3-configshell >= 1.1.fb25
 %endif
 %endif
 
-%if 0%{?rhel} == 7
+%if 0%{?rhel}
 BuildRequires: systemd
 %else
 BuildRequires: systemd-rpm-macros


ceph-iscsi.spec
```diff
@@ -75,7 +75,7 @@ Requires:       python3-configshell >= 1.1.fb25
 %endif
 %endif
 
-%if 0%{?rhel} == 7
+%if 0%{?rhel} <= 8
```
dillaman:
```suggestion
%if 0%{?rhel}
```

dillaman: If `rhel` is undefined (non-RHEL system) it will evaluate to 0 and pass the `<=` check.

mikechristie: Re-pushed with the fix.

Issue #100, opened 2019-05-31
Need some clarity on Manual installation


gvikram18:
Hi,

Under the installation section, It's mentioned that for the daemons, we gotta copy the files to the respective paths.

    <archive_root>/usr/lib/systemd/system/rbd-target-gw.service --> /lib/systemd/system
    <archive_root>/usr/lib/systemd/system/rbd-target-api.service --> /lib/systemd/system
    <archive_root>/usr/bin/rbd-target-gw --> /usr/bin
    <archive_root>/usr/bin/rbd-target-api --> /usr/bin
    <archive_root>/usr/bin/gwcli --> /usr/bin

But, looking at the repo, there's no `usr/bin` folder. Only `usr/lib/systemd/system/` exists.
The files mentioned to be placed at `/usr/bin` are rbd-target-gw, rbd-target-api, gwcli. Now, are these same files as the ones present in repo as rbd-target-gw.py, rbd-target-api.py and gwcli.py respectively? Need some clarity on this.

mikechristie:
Sorry about this. I missed this issue.

The docs are wrong. Just doing "python setup.py install" will install gwcli and the rbd-target-api and rbd-target-gw daemons to the correct places.

You do have to install the systemd unilt files still.

PR to fix the doc:

https://github.com/ceph/ceph-iscsi/pull/116

PR #116, opened 2019-06-28
Update manual install instructions

mikechristie:
gwcli, rbd-target-api and rbd-target-gw are installed by setup.py.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,18 +38,16 @@ python-configshell
 python-cryptography
 python-flask
 
-To install the python package that provides the application logic, run the
-provided setup.py script i.e. ```> python setup.py install```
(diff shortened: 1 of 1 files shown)


mikechristie:
Fixes GH issue:

https://github.com/ceph/ceph-iscsi/issues/100

mikechristie:
Giving myself a review comment. What about just replacing the info in README.md with a link to the online docs:

http://docs.ceph.com/docs/master/rbd/iscsi-target-cli/
http://docs.ceph.com/docs/master/rbd/iscsi-target-cli-manual-install/

like how the ceph-ansible doc does? We then only need to maintain one doc.

PR #118, opened 2019-06-29
Handle the case where FQDN is the same as hostname

dillaman:
The upgrade from config version 9 to 10 might result in the
removal of all gateways if the gateway FQDNs are the same as
the hostnames.

Signed-off-by: Jason Dillaman <dillaman@redhat.com>


ricardoasmarques:
@dillaman thanks for fixing this

PR #120, opened 2019-07-09
Validate controls min/max value

ricardoasmarques:
This PR will validate if control values meet the requirements described in PR https://github.com/ceph/ceph-iscsi-cli/pull/147:

Screenshot from 2019-07-09 14-31-23

Additionally, the limits for each control are now returned by the API so REST consumers like Ceph Dashboard can validate the inputs beforehand (https://github.com/ceph/ceph/pull/28942).

### DIFF
--- a/ceph_iscsi_config/settings.py
+++ b/ceph_iscsi_config/settings.py
@@ -24,6 +24,36 @@ def init():
 class Settings(object):
     LIO_YES_NO_SETTINGS = ["immediate_data", "initial_r2t"]
 
+    LIO_INT_SETTINGS_LIMITS = {
+        "cmdsn_depth": {
+            "min": 1, "max": 512},
+        "dataout_timeout": {
+            "min": 2, "max": 60},
+        "nopin_response_timeout": {
+            "min": 3, "max": 60},
+        "nopin_timeout": {
+            "min": 3, "max": 60},
+        "first_burst_length": {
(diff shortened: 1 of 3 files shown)


ricardoasmarques:
It would be great if we have the same kind of validation for LUN controls, but I don't know the limits for those controls.

@mikechristie can you tell me the min/max value for `hw_max_sectors`, `max_data_area_mb`, `osd_op_timeout` and `qfull_timeout`, so I can update this PR?

mikechristie:
The kernel limits are a little hokey in that the kernel let's you set whatever is passed in then it just rollover if the value is too large. So we could just set some sane values like:

// 10 minutes
qfull_timeout = 0 - 600

// Do we ever have someone use larger than the default object size of 4 MBs? For our testing we see perf drop after 2 MB so maybe cap at 4?

hw_max_sectors = 1 - 8192

// The following would be based on the kernels cmdsn limit of 512 times hw_max_sectors above of 4 MB for the extreme case of a single lun per target so we can fill the session.

max_data_area_mb = 1 - 2048

This is the ceph rados_osd_op_timeout and I am not sure if there is a hard code limit (I do not see it at least), so again maybe just set some sane value like:

// 10 minutes
osd_op_timeout = 0 - 600

ricardoasmarques:
@mikechristie Thanks, I've updated this PR accordingly.

PR #126, opened 2019-07-30
Fix spelling s/follwing/following/

sathieu:


sathieu:
Caught by lintian, the Debian static checker.

PR #128, opened 2019-07-31
Avoid flooding the log during partial upgrade

sathieu:
Every config update flood the log of all nodes.

### DIFF
--- a/ceph_iscsi_config/common.py
+++ b/ceph_iscsi_config/common.py
@@ -166,7 +166,7 @@ class Config(object):
 
         if self.config['version'] <= 2:
             self.add_item("groups", element_name=None, initial_value={})
-            self.update_item("version", element_name=None, element_value=3)
+            self.update_version(3)
 
         if self.config['version'] == 3:
             iqn = self.config['gateways']['iqn']
@@ -215,19 +215,19 @@ class Config(object):
             self.del_item('clients', None)
             self.del_item('groups', None)
             self.update_item("gateways", None, gateways)
-            self.update_item("version", None, 4)
+            self.update_version(4)
 
         if self.config['version'] == 4:
             for disk_id, disk in self.config['disks'].items():
                 disk['backstore'] = USER_RBD
                 self.update_item("disks", disk_id, disk)
-            self.update_item("version", None, 5)
+            self.update_version(5)
 
         if self.config['version'] == 5:
             for target_iqn, target in self.config['targets'].items():
                 target['acl_enabled'] = True
                 self.update_item("targets", target_iqn, target)
-            self.update_item("version", None, 6)
+            self.update_version(6)
 
         if self.config['version'] == 6:
             new_disks = {}
@@ -260,7 +260,7 @@ class Config(object):
                         new_group_disks[new_group_disk_id] = group_disk
                         group['disks'] = new_group_disks
                 self.update_item("targets", iqn, target)
-            self.update_item("version", None, 7)
+            self.update_version(7)
 
         if self.config['version'] == 7:
             if '/' in self.config['discovery_auth']['chap']:
@@ -306,7 +306,7 @@ class Config(object):
                     client['auth'].pop('chap_mutual', None)
 
                 self.update_item("targets", target_iqn, target)
-            self.update_item("version", None, 8)
+            self.update_version(8)
 
         if self.config['version'] == 8:
             for target_iqn, target in self.config['targets'].items():
@@ -314,7 +314,7 @@ class Config(object):
                     portal['portal_ip_addresses'] = [portal['portal_ip_address']]
                     portal.pop('portal_ip_address')
                 self.update_item("targets", target_iqn, target)
-            self.update_item("version", None, 9)
+            self.update_version(9)
 
         if self.config['version'] == 9 or self.config['version'] == 9.5:
             # temporary field to store the gateways already upgraded from v9 to v10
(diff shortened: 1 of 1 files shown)


ceph_iscsi_config/common.py
```diff
@@ -488,6 +488,11 @@ def del_item(self, cfg_type, element_name):
         txn = ConfigTransaction(cfg_type, element_name, 'delete')
         self.txn_list.append(txn)
 
+    def update_version(self, version):
+        if self.config.get('version', 0) == version:
+            return
+        self.update_item('version', None, version)
```
mikechristie: I'm not sure I understand what you are trying to fix.  It sounds like multiple nodes are running _upgrade_config. To me the patch then makes it look like they are running at the same time where initially they might all see version=N, but by the time we get to the update_item call, one of the nodes has done a commit already and updated the version to N+1. This patch would then prevent the other nodes from doing a commit. Is that right?

sathieu: I had a situation where only one node was upgraded. This flooded the log because other nodes were not.

The already upgraded node was looping in `# upgrade from v9 to v10 is still in progress, some gateways are not upgraded yet` (line 351) and setting version to the already set version.

mikechristie: Sorry for the late reply. I think when you marked it as resolved I missed that button when I went to check on it later thinking you never replied :)

I think for all but the 9.5 case, we will never hit the version check you are adding because we are always upgrading from N to N+1. And, on the swing around to handle the epoch change and refresh,  we will normally then hit this check:

```
    def _upgrade_config(self):
        if self.config['version'] >= Config.seed_config['version']:
            return
```

The 9.5 case is odd/special. I think we just want to add a check in that chunk:

```
if self.config['version'] == 9 or self.config['version'] == 9.5:

....


if self.config['version'] == 9:
     self.update_item("version", None, 9.5)
```

so the oddness for that case is not hidden. If it was common then I would say make it a function like you did.

mikechristie: I think we actually I need one more thing. For your patch and my idea about, the commit call still updates the epoch even though there are no changes, so we will still loop around over and over:


```
diff --git a/ceph_iscsi_config/common.py b/ceph_iscsi_config/common.py
index 56e6afc..1775962 100644
--- a/ceph_iscsi_config/common.py
+++ b/ceph_iscsi_config/common.py
@@ -164,9 +164,12 @@ class Config(object):
         if self.config['version'] >= Config.seed_config['version']:
             return
 
+        updated = False
+
         if self.config['version'] <= 2:
             self.add_item("groups", element_name=None, initial_value={})
             self.update_item("version", element_name=None, element_value=3)
+            updated = True
 
         if self.config['version'] == 3:
             iqn = self.config['gateways']['iqn']
@@ -216,18 +219,21 @@ class Config(object):
             self.del_item('groups', None)
             self.update_item("gateways", None, gateways)
             self.update_item("version", None, 4)
+            updated = True
 
         if self.config['version'] == 4:
             for disk_id, disk in self.config['disks'].items():
                 disk['backstore'] = USER_RBD
                 self.update_item("disks", disk_id, disk)
             self.update_item("version", None, 5)
+            updated = True
 
         if self.config['version'] == 5:
             for target_iqn, target in self.config['targets'].items():
                 target['acl_enabled'] = True
                 self.update_item("targets", target_iqn, target)
             self.update_item("version", None, 6)
+            updated = True
 
         if self.config['version'] == 6:
             new_disks = {}
@@ -261,6 +267,7 @@ class Config(object):
                         group['disks'] = new_group_disks
                 self.update_item("targets", iqn, target)
             self.update_item("version", None, 7)
+            updated = True
 
         if self.config['version'] == 7:
             if '/' in self.config['discovery_auth']['chap']:
@@ -307,6 +314,7 @@ class Config(object):
 
                 self.update_item("targets", target_iqn, target)
             self.update_item("version", None, 8)
+            updated = True
 
         if self.config['version'] == 8:
             for target_iqn, target in self.config['targets'].items():
@@ -315,6 +323,7 @@ class Config(object):
                     portal.pop('portal_ip_address')
                 self.update_item("targets", target_iqn, target)
             self.update_item("version", None, 9)
+            updated = True
 
         if self.config['version'] == 9 or self.config['version'] == 9.5:
             # temporary field to store the gateways already upgraded from v9 to v10
@@ -322,6 +331,7 @@ class Config(object):
             if not gateways_upgraded:
                 gateways_upgraded = []
                 self.add_item('gateways_upgraded', None, gateways_upgraded)
+                updated = True
             this_shortname = socket.gethostname().split('.')[0]
             this_fqdn = socket.getfqdn()
             if this_fqdn not in gateways_upgraded:
@@ -346,15 +356,20 @@ class Config(object):
                     self.update_item("disks", disk_id, disk)
                 gateways_upgraded.append(this_fqdn)
                 self.update_item("gateways_upgraded", None, gateways_upgraded)
+                updated = True
             if any(gateway_name not in gateways_upgraded
                    for gateway_name in self.config['gateways'].keys()):
                 # upgrade from v9 to v10 is still in progress, some gateways are not upgraded yet
-                self.update_item("version", None, 9.5)
+                if self.config['version'] == 9:
+                    self.update_item("version", None, 9.5)
+                    updated = True
             else:
                 self.del_item("gateways_upgraded", None)
                 self.update_item("version", None, 10)
+                updated = True
 
-        self.commit("retain")
+        if updated:
+            self.commit("retain")
 
     def init_config(self):
         try:
```

mikechristie: I made a PR here with this patch

https://github.com/ceph/ceph-iscsi/pull/144

mikechristie:
Fixed in

https://github.com/ceph/ceph-iscsi/pull/144

sathieu:
@mikechristie thanks

PR #132, opened 2019-08-25
Remove RPM kernel check to add support for other OS without YUM packa…

JKJameson:
…ge manager.

♥ Debian


rbd-target-api.py
```diff
@@ -2655,13 +2656,6 @@ def pre_reqs_errors():
     this_kernel = os_info[2].replace(".{}".format(this_arch), '')
     this_ver, this_rel = this_kernel.split('-', 1)
 
-    # use labelCompare from the rpm module to handle the comparison
-    if labelCompare(('1', this_ver, this_rel), ('1', k_vers, k_rel)) < 0:
```
dillaman: I'm fine w/ dropping the dependency on the `rpm` package (please also remove from the `.spec` and `tox.ini` files), but this now drops the kernel version checker all together. Perhaps just re-implement [1] within `ceph_iscsi_config/utils.py`

[1] https://pythonhosted.org/version_utils/_modules/version_utils/rpm.html#compare_evrs

JKJameson: It it naughty to just copy and paste the required functions into rbd-target-api.py ? I would add a block around the check, so if the OS is Debian, don't check the kernel package as it's built-in by default.

JKJameson: I changed the code to implement your suggested changes. I don't mind if we scratch that. It's untested and needs validation. Tell me what you prefer and we can roll with it. :)

JKJameson:
Let me know if you need anything else.

Issue #130, opened 2019-08-15
TPG level CHAP authentication and LUN id can't be configured


ddiss:
I'm trying to represent the following simple targetcli configuration in ceph-iscsi syntax, but two things appear to be unsupported:
- explicit LUN1 at a TPG level, instead of a client mapping with explicit LUN id
- TPG level CHAP authentication settings

```
# targetcli ls
  o- backstores ..................................................... [...]
...
  | o- user:rbd ...................................... [Storage Objects: 1] 
  |   o- rbd-lrbd_test ................ [rbd/lrbd_test (10.0GiB) activated]
  |     o- alua .......................................... [ALUA Groups: 1] 
  |       o- default_tg_pt_gp .............. [ALUA state: Active/optimized]
  o- iscsi .................................. [1-way disc auth, Targets: 1] 
  | o- iqn.2016-11.org.linux-iscsi.igw:stuff .................... [TPGs: 1]
  |   o- tpg1 ............................ [gen-acls, tpg-auth, 1-way auth]
  |     o- acls ................................................. [ACLs: 0] 
  |     o- luns ................................................. [LUNs: 1] 
  |     | o- lun1 ................. [user/rbd-lrbd_test (default_tg_pt_gp)]
  |     o- portals ........................................... [Portals: 1] 
  |       o- 192.168.1.1:3260 ........................................ [OK]
```

ddiss:
I'm happy to work on this, but would appreciate some input on how best to expose these options to the user. At the moment, TPGs appear to be hidden, with configuration instead taking place at a portal level.

mikechristie:
1. I am not sure what you mean by explicit LUN1. I mean I have no idea what that is wrt scsi, lio, and targetcli :) Can you maybe point me to the targetcli/rtslib code that implements it?

Is it SUSE specific by any chance?

Why LUN 1 and not LUN 0 btw? Are you using LUN 0 as the special LUN only?

2. Just to make sure I got this one,  you have a target with multiple tpgs, but then you want each tpg to have different CHAP settings? What is the user case  for this?

I think the reason the tpgs were hidden from the user was because for ESX you needed them to be in a specific order across all gateways so when the ALUA info is reported it looks the same through any port (some versions of ESX look at all the info in RTPG info and not the local one like defined in newer specs and like how linux does it). When going from targetcli to gwcli we did not allow the user to config tpgs, because they kept messing things up when we were doing the targetcli based deployments.

[...]

PR #133, opened 2019-08-26
Target level CHAP authentication and TPG lun_id persistence

ricardoasmarques:
This PR contains 2 different commits to fix both topics discussed on issue https://github.com/ceph/ceph-iscsi/issues/130

**1. Target level CHAP authentication**
It's now possible to define CHAP/MUTUAL_CHAP authentication at the target level (same credentials are applied to all TPGs).
When using ACL, those credentials will be inherited and can be overridden at the client level.

**2. Persist TPG lun_id**
Instead of relying on storage object path to generate it, TPG `lun_id` will now be persisted in the `gateway.conf` ensuring that the TPG `lun_id` is the same in all gateways, and will not change across restarts.

Fixes https://github.com/ceph/ceph-iscsi/issues/130

### DIFF
--- a/ceph_iscsi_config/client.py
+++ b/ceph_iscsi_config/client.py
@@ -307,7 +307,8 @@ class GWClient(GWObject):
 
             client.manage('present')  # ensure the client exists
 
-    def try_disable_auth(self, tpg):
+    @staticmethod
+    def try_disable_auth(tpg):
         """
         Disable authentication (enable ACL mode) if this is the last CHAP user.
 
@@ -318,6 +319,9 @@ class GWClient(GWObject):
         for client in tpg.node_acls:
             if client.chap_userid or client.chap_password:
                 return
+
+        if tpg.chap_userid or tpg.chap_password:
+            return
 
         tpg.set_attribute('authentication', '0')
 
@@ -371,7 +375,7 @@ class GWClient(GWObject):
             if auth_enabled:
                 self.tpg.set_attribute('authentication', '1')
             else:
-                self.try_disable_auth(self.tpg)
+                GWClient.try_disable_auth(self.tpg)
 
             self.logger.debug("Updating config object meta data")
             encryption_enabled = encryption_available()
@@ -485,7 +489,7 @@ class GWClient(GWObject):
 
         try:
             self.acl.delete()
-            self.try_disable_auth(self.tpg)
+            GWClient.try_disable_auth(self.tpg)
             self.change_count += 1
             self.logger.info("(Client.delete) deleted NodeACL for "
                              "{}".format(self.iqn))
--- a/ceph_iscsi_config/common.py
+++ b/ceph_iscsi_config/common.py
@@ -364,6 +364,12 @@ class Config(object):
                     'mutual_password': '',
                     'mutual_password_encryption_enabled': False
                 }
+                disks = {}
+                for disk_index, disk in enumerate(sorted(target['disks'])):
+                    disks[disk] = {
+                        'lun_id': disk_index
+                    }
+                target['disks'] = disks
                 self.update_item("targets", target_iqn, target)
             self.update_item("version", None, 11)
 
--- a/ceph_iscsi_config/common.py
+++ b/ceph_iscsi_config/common.py
@@ -56,7 +56,7 @@ class Config(object):
                                       'mutual_username': '',
                                       'mutual_password': '',
                                       'mutual_password_encryption_enabled': False},
-                   "version": 10,
+                   "version": 11,
                    "epoch": 0,
                    "created": '',
                    "updated": ''
@@ -353,6 +353,19 @@ class Config(object):
             else:
                 self.del_item("gateways_upgraded", None)
                 self.update_item("version", None, 10)
+
+        if self.config['version'] == 10:
+            for target_iqn, target in self.config['targets'].items():
+                target['auth'] = {
+                    'username': '',
+                    'password': '',
+                    'password_encryption_enabled': False,
+                    'mutual_username': '',
+                    'mutual_password': '',
+                    'mutual_password_encryption_enabled': False
+                }
+                self.update_item("targets", target_iqn, target)
+            self.update_item("version", None, 11)
 
         self.commit("retain")
 
--- a/ceph_iscsi_config/gateway.py
+++ b/ceph_iscsi_config/gateway.py
@@ -356,7 +356,7 @@ class CephiSCSIGateway(object):
                 if not target_config['portals']:
                     # Last gw for the target so delete everything that lives
                     # under the tpg in LIO since we can't create it
-                    target_config['disks'] = []
+                    target_config['disks'] = {}
                     target_config['clients'] = {}
                     target_config['controls'] = {}
                     target_config['groups'] = {}
--- a/ceph_iscsi_config/lun.py
+++ b/ceph_iscsi_config/lun.py
@@ -406,7 +406,7 @@ class LUN(GWObject):
             # by using the allocating host we ensure the delete is not
             # issue by several hosts when initiated through ansible
 
-            target_config['disks'].remove(self.config_key)
+            target_config['disks'].pop(self.config_key)
             self.config.update_item("targets", target_iqn, target_config)
 
             # determine which host was the path owner
@@ -430,13 +430,25 @@ class LUN(GWObject):
 
             self.config.commit()
 
+    def _get_next_lun_id(self, target_disks):
+        lun_ids_in_use = [t['lun_id'] for t in target_disks.values()]
+        lun_id_candidate = 0
+        while lun_id_candidate in lun_ids_in_use:
+            lun_id_candidate += 1
+        return lun_id_candidate
+
     def map_lun(self, gateway, owner, disk):
         target_config = self.config.config['targets'][gateway.iqn]
         disk_metadata = self.config.config['disks'][disk]
         disk_metadata['owner'] = owner
         self.config.update_item("disks", disk, disk_metadata)
 
-        target_config['disks'].append(disk)
+        target_disk_config = target_config['disks'].get(disk)
+        if not target_disk_config:
+            lun_id = self._get_next_lun_id(target_config['disks'])
+            target_config['disks'][disk] = {
+                'lun_id': lun_id
+            }
         self.config.update_item("targets", gateway.iqn, target_config)
 
         gateway_dict = self.config.config['gateways'][owner]
@@ -447,7 +459,7 @@ class LUN(GWObject):
         if self.error:
             raise CephiSCSIError(self.error_msg)
 
-        gateway.map_lun(self.config, so)
+        gateway.map_lun(self.config, so, target_config['disks'][disk])
         if gateway.error:
             raise CephiSCSIError(gateway.error_msg)
 
@@ -509,7 +521,7 @@ class LUN(GWObject):
             # Add the mapping for the lun to ensure the block device is
             # present on all TPG's
             gateway = GWTarget(self.logger, target_iqn, ip_list)
-            gateway.map_lun(self.config, so)
+            gateway.map_lun(self.config, so, target['disks'][self.config_key])
             if gateway.error:
                 raise CephiSCSIError("LUN mapping failed - {}".format(gateway.error_msg))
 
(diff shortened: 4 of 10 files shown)


ddiss:
As discussed, I'm a little confused by the  need for inheritance of TPG level CHAP settings into the client ACLs. I'd like to understand why we can't rely on the default LIO behaviour here, so will do some further investigation.

ricardoasmarques:
> As discussed, I'm a little confused by the need for inheritance of TPG level CHAP settings into the client ACLs. I'd like to understand why we can't rely on the default LIO behaviour here, so will do some further investigation.

Setting auth credentials at target-level and have those credentials applied to all ACL clients seemed to me like a useful behavior from the usability point of view but if we conclude that it's not a good idea I can change it.

@mikechristie do you have an opinion about this?

@ddiss thanks for looking into the default LIO behavior, let us know when you have any conclusion.

ddiss:
> As discussed, I'm a little confused by the need for inheritance of TPG level CHAP settings into the client ACLs. I'd like to understand why we can't rely on the default LIO behaviour here, so will do some further investigation.

I had a play with this on mainline kernel and LIO behaved as expected:

- ${tpgt}/attrib/authentication=1, ${tpgt}/attrib/generate_node_acls=1, ${tpgt}/auth set and ${tpgt}/acls/${initiator}/auth set
  - initiators with explicit acls/${initiator} entries use those auth settings, initiators without explicit acls use TPG auth settings
- ${tpgt}/attrib/authentication=1, ${tpgt}/attrib/generate_node_acls=0, ${tpgt}/auth set and ${tpgt}/acls/${initiator}/auth set
  - only initiators with explicit acls can login using acl auth settings

@ricardoasmarques you mentioned that you were seeing some unexpected behaviour when not performing inheritance within ceph-iscsi. Could you elaborate a bit? It'd be good if you could also share your configfs settings for this case.

ricardoasmarques:
@ddiss Note that your first example is not valid when using `ceph-iscsi` because `generate_node_acls` is automatically set to `0` when you add an ACL/initiator from `ceph-iscsi` (and you cannot set it back to `1` unless you remove all ACLs/initiators).

The scenario I was trying to configure using `ceph-iscsi` was:
- set target-level chap auth
- add an initiator to the target
- _(I was surprised that the initiator is not able to connect using credentials defined at the target level, thats why I've implemented the "inheritence")_

This is the LIO configuration before I've implemented the "inheritence":

- ${tpgt}/attrib/authentication=1
- ${tpgt}/attrib/generate_node_acls=0
- ${tpgt}/auth set
- ${tpgt}/acls/${initiator}/auth **NOT** set
- _(initiator **cannot** connect using `${tpgt}/auth` credentials)_

With the "inheritence" that I've implemented LIO will be:

- ${tpgt}/attrib/authentication=1
- ${tpgt}/attrib/generate_node_acls=0
- ${tpgt}/auth set
- ${tpgt}/acls/${initiator}/auth set (with credentials from ${tpgt}/auth)
- _(initiator **can** connect using `${tpgt}/auth` credentials)_

ddiss:
> @ddiss Note that your first example is not valid when using `ceph-iscsi` because `generate_node_acls` is automatically set to `0` when you add an ACL/initiator from `ceph-iscsi` (and you cannot set it back to `1` unless you remove all ACLs/initiators).

Okay, I guess that makes sense.

> 
> The scenario I was trying to configure using `ceph-iscsi` was:
> 
>     * set target-level chap auth
> 
>     * add an initiator to the target
> 
>     * _(I was surprised that the initiator is not able to connect using credentials defined at the target level, thats why I've implemented the "inheritence")_
> 
> 
> This is the LIO configuration before I've implemented the "inheritence":
> 
>     * ${tpgt}/attrib/authentication=1
> 
>     * ${tpgt}/attrib/generate_node_acls=0
> 
>     * ${tpgt}/auth set
> 
>     * ${tpgt}/acls/${initiator}/auth **NOT** set
> 
>     * _(initiator **cannot** connect using `${tpgt}/auth` credentials)_

With  ${tpgt}/attrib/authentication=1 configured, LIO doesn't do any fallback from ${tpgt}/acls/${initiator}/auth to the TPG auth settings. acls must be configured with auth settings.
I wasn't aware that ceph-iscsi allowed for "client" configurations to be present without auth settings.


> With the "inheritence" that I've implemented LIO will be:
> 
>     * ${tpgt}/attrib/authentication=1
> 
>     * ${tpgt}/attrib/generate_node_acls=0
> 
>     * ${tpgt}/auth set
> 
>     * ${tpgt}/acls/${initiator}/auth set (with credentials from ${tpgt}/auth)
> 
>     * _(initiator **can** connect using `${tpgt}/auth` credentials)_

If a "client" configuration is present with explicit auth settings will the TPG auth settings be ignored or inherited (overwriting the explicit settings)?

ricardoasmarques:
> With ${tpgt}/attrib/authentication=1 configured, LIO doesn't do any fallback from ${tpgt}/acls/${initiator}/auth to the TPG auth settings. acls must be configured with auth settings.
> I wasn't aware that ceph-iscsi allowed for "client" configurations to be present without auth settings.

@ddiss  Yes, `ceph-iscsi` allows using ACL without auth credentials.

> If a "client" configuration is present with explicit auth settings will the TPG auth settings be ignored or inherited (overwriting the explicit settings)?

@ddiss Explicit "client-level" auth will override the "target-level" auth, for example, doing the following in `ceph-iscsi`:

- set target-level auth (**user1/pass1**)
- add initiator-1
- add initiator-2
- set client-level auth for initiator-1 (**user2/pass2**)

Will result in the following LIO configuration:

- ${tpgt}/attrib/authentication=1
- ${tpgt}/attrib/generate_node_acls=0
- ${tpgt}/auth set to **user1/pass1**
- ${tpgt}/acls/${initiator-1}/auth set to **user2/pass2**
- ${tpgt}/acls/${initiator-2}/auth set to **user1/pass1**

ddiss:
> > If a "client" configuration is present with explicit auth settings will the TPG auth settings be ignored or inherited (overwriting the explicit settings)?
> 
> @ddiss Explicit "client-level" auth will override the "target-level" auth, for example, doing the following in `ceph-iscsi`:
> 
>     * set target-level auth (**user1/pass1**)
> 
>     * add initiator-1
> 
>     * add initiator-2
> 
>     * set client-level auth for initiator-1 (**user2/pass2**)
> 
> 
> Will result in the following LIO configuration:
> 
>     * ${tpgt}/attrib/authentication=1
> 
>     * ${tpgt}/attrib/generate_node_acls=0
> 
>     * ${tpgt}/auth set to **user1/pass1**
> 
>     * ${tpgt}/acls/${initiator-1}/auth set to **user2/pass2**
> 
>     * ${tpgt}/acls/${initiator-2}/auth set to **user1/pass1**

Thanks for the explanation. If the target-level auth is removed after this point, does initiator-2 retain the inherited username / password? Given that ACL auth is either enabled or disabled for all initiators, I can't see what else would make sense here.

I'd be interested to hear @mikechristie 's thoughts on whether the ACL->TPG auth fallback should be handled in kernel.

mikechristie:
Give me a day to think about it some more.

Just some background, because I want to make sure we are not carrying over some funky behavior due to my misunderstanding of how LIO works.

My understanding was that you cannot have a TPG with 2 ACLs where one uses CHAP and one uses initiator name based auth. There is a funky way to almost enable it by setting the AuthMethod directly, but it results in some buggy cases.

So with that assumption, we had allowed you to create ACLs with CHAP and without but it currently can leave the system in a weird state where some work and some do not and we just document the user has to pick one (gwcli lists the auth state as misconfigured). This is basically just to let the user switch between CHAP and initiator name based ACLs or the reverse while they set up.

My goal was to fix up the kernel so it allowed this type of mixing and what I think you describe as auth fallback. Give me a couple days to try and recover it. I think I hit several issues and just didn't have time to complete it and my workstation died first :) but I have the drive still.

ricardoasmarques:
> > > If a "client" configuration is present with explicit auth settings will the TPG auth settings be ignored or inherited (overwriting the explicit settings)?
> > 
> > 
> > @ddiss Explicit "client-level" auth will override the "target-level" auth, for example, doing the following in `ceph-iscsi`:
> > ```
> > * set target-level auth (**user1/pass1**)
> > 
> > * add initiator-1
> > 
> > * add initiator-2
> > 
> > * set client-level auth for initiator-1 (**user2/pass2**)
> > ```
> > 
> > 
> > Will result in the following LIO configuration:
> > ```
> > * ${tpgt}/attrib/authentication=1
> > 
> > * ${tpgt}/attrib/generate_node_acls=0
> > 
> > * ${tpgt}/auth set to **user1/pass1**
> > 
> > * ${tpgt}/acls/${initiator-1}/auth set to **user2/pass2**
> > 
> > * ${tpgt}/acls/${initiator-2}/auth set to **user1/pass1**
> > ```
> 
> Thanks for the explanation. If the target-level auth is removed after this point, does initiator-2 retain the inherited username / password? Given that ACL auth is either enabled or disabled for all initiators, I can't see what else would make sense here.

In the current implementation of this PR, you will end up with the following LIO implementation:

* ${tpgt}/attrib/authentication=1

* ${tpgt}/attrib/generate_node_acls=0

* ${tpgt}/auth **NOT** set
 
* ${tpgt}/acls/${initiator-1}/auth set to **user2/pass2**

* ${tpgt}/acls/${initiator-2}/auth **NOT** set

As @mikechristie said, this will be considered a "misconfigured" and "initiator-2" will not be able to connect until auth is set for "initiator-2" or auth is removed from "initiator-1".

mikechristie:
I can't find my old patches, but I think we need a new setting on the ACL that indicates if we do:

1. CHAP using the ACL values.
2. Initiator name based ACL (just match initiator name on the ACL).
3. Fall back to the TPG level authentication values.

In iscsi_target_handle_csg_zero we would then check this new setting instead of only going by the tpg authentication setting.

ddiss:
> I can't find my old patches, but I think we need a new setting on the ACL that indicates if we do:
> 
>     CHAP using the ACL values.
>     Initiator name based ACL (just match initiator name on the ACL).
>     Fall back to the TPG level authentication values.
> 
> In iscsi_target_handle_csg_zero we would then check this new setting instead of only going by the tpg authentication setting.

Sounds reasonable to me. I'll put something together to do this.

ricardoasmarques:
@ddiss @mikechristie Any chance we can review/merge this PR before implementing the kernel changes, or do we have to block it until then?

Maybe we can use the "manual fallback to target auth" implemented in this PR for now, and adapt it later when kernel supports the fallback at LIO level. Or we can remove the "auth fallback" from this PR. Thoughts?

mikechristie:
The lun_id patch should definitely go in before the auth stuff since it's not related and I think it fixes a bug.

For the CHAP part, I am ok with any option as long as we can do it in a way that does not break setups in the future or we should at least have a upgrade path.

For the CHAP behavior, I am ok with the behavior implemented in your patch. If we want to discuss it some more and/or make sure we can make the kernel interface do what we want because that code can get quirky, I would suggest to break it up into 2 PRs so the lun_id stuff go in first.

ricardoasmarques:
@mikechristie to avoid having to wait for kernel changes, and to avoid having to remove the code that implements the "auth fallback" in `ceph-iscsi` later, I've updated this PR with a different approach.

With my latest changes `ceph-iscsi` will **NOT** allow mixing TPG/target auth with ACL auth.

This means that:
- if TPG/target auth is being used, then the user is not allowed to enable ACL / add clients
- if ACL auth is enabled, then the user is not able to set TPG/target auth

If in the future we really have the need of mixing both, then we can follow up the `kernel`+`ceph-iscsi` discussion later, and have a dedicated PR just for that.

@mikechristie @ddiss if you agree with this "simpler" approach, then this PR is now ready for review. Thanks.

mikechristie:
That seems ok to me.

mikechristie:
@ddiss 

I didn't see any objections from you and see you built some other patches on @ricardoasmarques's set here, so I am going to merge it up.

Feel free to just send a PR for whatever changes you want in the future.

mikechristie:
Just adding a note.

This exposes an issue where one node will upgrade to 9.5 only but the code now assumes 11, so if the other nodes do not upgrade fast enough, you will hit a crash when trying to access the new config object fields during initialization. That will be fixed in a updated

https://github.com/ceph/ceph-iscsi/pull/144

which will fix the loop issue and this one.

PR #143, opened 2019-09-19
ceph-iscsi settings cleanup

mikechristie:
The lio/ceph-iscsi settings values, limits, and definitions are spread all over. This moves it all to a new file and tries to remove a lot of the duplicated loops/checks/conversion/normalization code.

### DIFF
--- a/ceph_iscsi_config/gateway_setting.py
+++ b/ceph_iscsi_config/gateway_setting.py
@@ -1,0 +1,202 @@
+import logging
+
+
+def convert_str_to_bool(value):
+    """
+    Convert true/false/yes/no/1/0 to boolean
+    """
+
+    if isinstance(value, bool):
+        return value
+
+    value = str(value).lower()
+    if value in ['1', 'true', 'yes']:
+        return True
+    elif value in ['0', 'false', 'no']:
+        return False
+    raise ValueError(value)
+
+
+class Setting(object):
+    def __init__(self, name, type_str, def_val):
+        self.name = name
+        self.type_str = type_str
+        self.def_val = def_val
+
+    def __contains__(self, key):
+        return key == self.def_val
+
+
+class BoolSetting(Setting):
+    def __init__(self, name, def_val):
+        super(BoolSetting, self).__init__(name, "bool", def_val)
+
+    def to_str(self, norm_val):
+        if norm_val:
+            return "true"
+        else:
+            return "false"
+
+    def normalize(self, raw_val):
+        try:
+            # for compat we also support Yes/No and 1/0
+            return convert_str_to_bool(raw_val)
+        except ValueError:
+            raise ValueError("expected true or false for {}".format(self.name))
+
+
+class LIOBoolSetting(BoolSetting):
+    def __init__(self, name, def_val):
+        super(LIOBoolSetting, self).__init__(name, def_val)
+
+    def to_str(self, norm_val):
+        if norm_val:
+            return "yes"
+        else:
+            return "no"
+
+    def normalize(self, raw_val):
+        try:
+            # for compat we also support True/False and 1/0
+            return convert_str_to_bool(raw_val)
+        except ValueError:
+            raise ValueError("expected yes or no for {}".format(self.name))
+
+
+class ListSetting(Setting):
+    def __init__(self, name, def_val):
+        super(ListSetting, self).__init__(name, "list", def_val)
+
+    def to_str(self, norm_val):
+        return str(norm_val)
+
+    def normalize(self, raw_val):
+        return raw_val.split(',') if raw_val else []
+
+
+class StrSetting(Setting):
+    def __init__(self, name, def_val):
+        super(StrSetting, self).__init__(name, "str", def_val)
+
+    def to_str(self, norm_val):
+        return str(norm_val)
+
+    def normalize(self, raw_val):
+        return str(raw_val)
+
+
+class IntSetting(Setting):
+    def __init__(self, name, min_val, max_val, def_val):
+        self.min_val = min_val
+        self.max_val = max_val
+        super(IntSetting, self).__init__(name, "int", def_val)
+
+    def to_str(self, norm_val):
+        return str(norm_val)
+
+    def normalize(self, raw_val):
+        try:
+            val = int(raw_val)
+        except ValueError:
+            raise ValueError("expected integer for {}".format(self.name))
+
+        if val < self.min_val:
+            raise ValueError("expected integer >= {} for {}".
+                             format(self.min_val, self.name))
+        if val > self.max_val:
+            raise ValueError("expected integer <= {} for {}".
+                             format(self.max_val, self.name))
+        return val
+
+
+class EnumSetting(Setting):
+    def __init__(self, name, valid_vals, def_val):
+        if len(valid_vals) == 0:
+            raise ValueError("Invalid enum. There must be at least one valid value.")
+
+        valid_type = type(valid_vals[0])
+        if valid_type is not int and valid_type is not str:
+            raise ValueError("Invalid enum. Items must be str or int. Got {}".
+                             format(valid_type))
+
+        for i in valid_vals:
+            if valid_type != type(i):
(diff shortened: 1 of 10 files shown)


rbd-target-api.py
```diff
@@ -2361,44 +2387,18 @@ def get_settings():
     Examples:
     curl --insecure --user admin:admin -X GET https://192.168.122.69:5000/api/settings
     """
-    type_int = "int"
-    type_bool = "bool"
-
```
mikechristie: Hey @ricardoasmarques 

I noticed one small possible future issue with the patch that added the types. We seem to have a mix of 2 kinds of bools.

1. LIO based bools. These are iscsi rfc values and use yes/no like defined in that spec and used in the kernel. immediate_data is an example of this type.

2. I guess this would be called ceph-iscsi bools. These are the standard true/false values. api_secure is an example of this kind.

Currently, get_settings only reports about type 1. For this we report type=bool, and then for the values that are reported we use Yes/No. If we ever add support for type 2, would we report type=bool but and for the values would we use True/False?

Note that the PUT api calls can handle true/false, yes/no and 1/0, so we can handle anything. But I was just wondering if for your use case, you want me to separate the 2 types? If so, we could do

1. type=lio_bool (yes/no)
2 type=bool (true/false)

or

bool is always yes/no

or keep what we have:

bool= yes/no or true/false

?

Let me know and I can fix that in this patchset if you want.

ricardoasmarques: From the point of view of an external REST API consumer, I think I would prefer to don't have to know about that special type and treat everything as real True/False booleans (and `ceph-iscsi` would do all the conversions internally).

But looking into the output of `/api/settings`:
```
"target_controls_limits": {
        ...
        "immediate_data": {
            "type": "bool"
        },
        ....
    },
    "target_default_controls": {
        ...
        "immediate_data": "Yes",
        ...
    },
```
I see that it would require to return `True` instead of `"Yes"` in the default values.

**Do you think it makes sense to convert the default value to a True/False boolean?** 
If yes, then having a single "bool" type seems good to me.
Otherwise, having a special `lio_bool` seems more correct because it will help the consumer on how to process the default "Yes" value.

ceph_iscsi_config/gateway_setting.py
```diff
@@ -0,0 +1,167 @@
...
+    "priv_key": StrSetting("priv_key", 'iscsi-gateway.key'),
+    "pub_key": StrSetting("pub_key", 'iscsi-gateway-pub.key'),
+    "prometheus_exporter": BoolSetting("prometheus_exporter", True),
+    "prometheus_port": IntSetting("prometheus_port", 1, 65535, 9287),
+    "prometheus_host": StrSetting("prometheus_host", "::"),
+    "logger_level": IntSetting("logger_level", logging.DEBUG, logging.CRITICAL,
+                               logging.DEBUG),
+    # TODO: This is under sys for compat. It is not settable per device/backend
+    # type yet.
+    "alua_failover_type": StrSetting("alua_failover_type", "implicit")}
```
ricardoasmarques: What about having an `EnumSetting` that knows about the list of possible values (and can validate the value based on that)?

Something like:
```
EnumSetting("alua_failover_type", "implicit", ["implicit", "other_valid_value"])
```

rbd-target-api.py
mikechristie: Ok. I kept bool and made it return true/false, so it is consistent.

yes/no will be used when interacting with lio internally and then we also continue to allow tools/users to pass in yes/no for iscsi RFC values for compatibility.

Let me know if I misunderstood you. My allergies are really bad and so reading is hurting my eyes.

ceph_iscsi_config/gateway_setting.py
mikechristie: Nice. Done.

rbd-target-api.py
```diff
@@ -2352,6 +2353,23 @@ def hostgroup(target_iqn, group_name):
...
+    limits = {}
+
+    for k, setting in def_settings.items():
+        # Return normalized value to match get_config()'s format
+        defaults[k] = getattr(settings.config, k)
+
+        if isinstance(setting, IntSetting):
+            limits[k] = {'min': setting.min_val, 'max': setting.max_val,
+                         'type': setting.type_str}
+        else:
```
ricardoasmarques: We should return the valid values for the `enum` type, something equivalent to:

```
elif isinstance(setting, EnumSetting):
    limits[k] = {'values': setting.valid_vals,
                 'type': setting.type_str}
```

ceph_iscsi_config/gateway_setting.py
```diff
@@ -0,0 +1,185 @@
...
+            raise ValueError("expected integer >= {} for {}".
+                             format(self.min_val, self.name))
+        if val > self.max_val:
+            raise ValueError("expected integer <= {} for {}".
+                             format(self.max_val, self.name))
+        return val
+
+
+class EnumSetting(Setting):
+    def __init__(self, name, def_val, valid_vals):
```
ricardoasmarques: Nit: I've noticed that `def_val` is the last arg in `IntSetting` and the first in the `EnumSetting`, does it make sense to uniformize for better readability?

This is what we have now:
```
class IntSetting(Setting):
    def __init__(self, name, min_val, max_val, def_val):
class EnumSetting(Setting):
    def __init__(self, name, def_val, valid_vals):
```
I was expecting something like:
```
class IntSetting(Setting):
    def __init__(self, name, def_val, min_val, max_val):
class EnumSetting(Setting):
    def __init__(self, name, def_val, valid_vals):
```
or 
```
class IntSetting(Setting):
    def __init__(self, name, min_val, max_val, def_val):
class EnumSetting(Setting):
    def __init__(self, name, valid_vals, def_val):
```

mikechristie: That is a bug.

ceph_iscsi_config/gateway_setting.py
```diff
@@ -0,0 +1,185 @@
...
+        self.valid_vals = valid_vals
+        super(EnumSetting, self).__init__(name, "enum", def_val)
+
+    def to_str(self, norm_val):
+        return str(norm_val)
+
+    def normalize(self, raw_val):
+        norm_val = str(raw_val)
```
ricardoasmarques: @mikechristie do we really need this conversion to `string`?

This is preventing me to use the enum type with `integers`, my use case is the following:

`"block_size": EnumSetting("block_size", 512, [512, 1024, 2048, 4096]),`

This scenario is throwing a validation error because, for instance, `int list [512, 1024, 2048, 4096]` does not contain `string "512"`.

Changing this method to:
```
if raw_val not in self.valid_vals:
    raise ValueError("expected {}".format(self.valid_vals))

return raw_val
```

solves my issue.

Or do you see a better way to handle both `enum of strings` and `enum of integers`?

ceph_iscsi_config/gateway_setting.py
```diff
@@ -0,0 +1,185 @@
...
+
+        if norm_val not in self.valid_vals:
+            raise ValueError("expected {}".format(self.valid_vals))
```
ricardoasmarques: Nit: Would `ValueError("expected {}, found {}".format(self.valid_vals, norm_val))` be more useful to the user?

ceph_iscsi_config/gateway_setting.py
mikechristie: Ah yeah, that is much better. I only had the one string enum example and hard coded it for that which is dumb on my part. Will update with your change.

ceph_iscsi_config/gateway_setting.py
mikechristie: Yes.

ceph_iscsi_config/gateway_setting.py
mikechristie: I had to modify this a little more, because the config file parser is going to give us everything in strings. So there are some extra checks to put the passed in value in the format of the defined valid values.

rbd-target-api.py
mikechristie: Fixed.

PR #144, opened 2019-09-19
Fix 9 to 10 upgrade

mikechristie:
[edit: updated description to indicate this also fixes issue 2]

This fixes 2 bugs:

1. While we are in version 9.5, the config watch would detect the
version being updated to 9.5 then swing back on us. We only check for
full version upgrades so we would end up updating the version again and
looping like this until the other nodes were updated.

2. If upgrading from 9 to version > 10, the new versions added fields
that were then later refefenced when setting up the target. If all the
gws did not upgrade quickly enough, like with a rolling upgrade, then
the config watch might not have detected the config change and updated
us to 11 and added the new fields. We would then crash.

This patch skips the intermediate version 9.5 and checks if the node is
in the gateways_upgraded list or not to determine if the node needs to
update it's gw name fields still.

This is very fragile since it relies on no future upgrades needing the
updated fqdn based config values. However, I am not sure if it's worth
it to add code to sync our state at this point if we do not need it. If
we do end up needing this, then it might be better to have users upgrade
to a version 10, then from there upgrade to the current version.

### DIFF
--- a/ceph_iscsi_config/common.py
+++ b/ceph_iscsi_config/common.py
@@ -160,8 +160,25 @@ class Config(object):
 
         return cfg_dict
 
+    def needs_hostname_update(self):
+        if self.config['version'] == 9:
+            # No gateway has been updated yet.
+            return True
+
+        updated = self.config.get('gateways_upgraded')
+        if updated is None:
+            # Everything has been updated or we are < 9
+            return False
+
+        if socket.getfqdn() in updated:
+            return False
+
+        return True
+
     def _upgrade_config(self):
-        if self.config['version'] >= Config.seed_config['version']:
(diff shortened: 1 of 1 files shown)


mikechristie:
This fixes the issue described here:

https://github.com/ceph/ceph-iscsi/pull/128

mikechristie:
@ricardoasmarques / @dillaman 

I updated the patch in a way that you will want to review in case you want to take back your approval :) It is super hacky, but I am not sure what else to do other than force users to upgrade from version < 9 to version=10 on all nodes, then perform a second upgrade to version > 10.

ricardoasmarques:
@mikechristie your latest patch LGTM. Thanks.

PR #147, opened 2019-09-26
Add python requests

mikechristie:
Fix

Sep 25 12:12:29 iscsi-gw0 rbd-target-api[9959]: import requests
Sep 25 12:12:29 iscsi-gw0 rbd-target-api[9959]: ImportError: No module named requests

python-requests used to be brought in my ceph-common, so we never
noticed we were missing out dep on it. This adds it to the spec

### DIFF
--- a/ceph-iscsi.spec
+++ b/ceph-iscsi.spec
@@ -50,8 +50,10 @@ Requires:       python-flask >= 0.10.1
 Requires:       python-configshell >= 1.1.fb25
 %if 0%{?rhel} == 7
 Requires:       pyOpenSSL
+Requires:       python-requests
 %else
 Requires:       python-pyOpenSSL
+Requires:       python2-requests
 %endif
 %else
 BuildRequires:  python3-devel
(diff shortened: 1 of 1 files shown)


ceph-iscsi.spec
```diff
@@ -48,6 +48,7 @@ Requires:       python-rtslib >= 2.1.fb68
 Requires:       python-cryptography
 Requires:       python-flask >= 0.10.1
 Requires:       python-configshell >= 1.1.fb25
+Requires:       python2-requests
```
dillaman: Nit: is this the name of the package under SUSE? Google implies it's just `python-requests`

mikechristie: I'm not sure now, but I think it's python2-requests. I had looked it up here:

https://software.opensuse.org/package/python2-requests
but I do see:
https://software.opensuse.org/package/python-requests

For both, it does not show a SLES rpm. I just see the open suse one.

For opensuse, when I did a zypper install python2-requests it installed that rpm.

I just tried "zypper install python-requests" and the rpm that gets installed is python2-requests.

Adding @ricardoasmarques  just in case.

mikechristie: Dang, but I did mess up the RHEL 7 rpm. It is named python-requests :)

PR #150, opened 2019-10-17
Allow explicit set LUN number and WWN

ricardoasmarques:
This PR will allow the user to optionally specify the **LUN number** and **WWN**.

Note that if not explicitly set, then the LUN number and WWN will be automatically generated.

With this change `"ceph-iscsi" + "Ceph Dashboard"` will reach feature parity with `"lrbd" + "openATTIC"`. 

Just for reference, here is a print screen from the openATTIC:
Screenshot from 2019-10-17 09-25-52

Also, this PR adds a new `api_version` field to the settings endpoint response so consumers of the API can have more control on how to consume the API.

This `api_version` should be incremented every time the API changes.

### DIFF
--- a/ceph_iscsi_config/client.py
+++ b/ceph_iscsi_config/client.py
@@ -52,22 +52,6 @@ class GWClient(GWObject):
         self.target_iqn = target_iqn
         self.lun_lookup = {}        # only used for hostgroup based definitions
         self.requested_images = []
-
-        # image_list is normally a list of strings (pool/image_name) but
-        # group processing forces a specific lun id allocation to masked disks
-        # in this scenario the image list is a tuple
-        if image_list:
-
-            if isinstance(image_list[0], tuple):
-                # tuple format ('disk_name', {'lun_id': 0})...
-                for disk_item in image_list:
-                    disk_name = disk_item[0]
-                    lun_id = disk_item[1].get('lun_id')
-                    self.requested_images.append(disk_name)
-                    self.lun_lookup[disk_name] = lun_id
-            else:
-                self.requested_images = image_list
-
         self.username = username
         self.password = password
         self.mutual_username = mutual_username
@@ -113,6 +97,34 @@ class GWClient(GWObject):
         except CephiSCSIError as err:
             self.error = True
             self.error_msg = err
+
+        # image_list is normally a list of strings (pool/image_name) but
+        # group processing forces a specific lun id allocation to masked disks
+        # in this scenario the image list is a tuple
+        if image_list:
+
+            if isinstance(image_list[0], tuple):
+                # tuple format ('disk_name', {'lun_id': 0})...
+                for disk_item in image_list:
+                    disk_name = disk_item[0]
+                    lun_id = disk_item[1].get('lun_id')
+                    self.requested_images.append(disk_name)
+                    self.lun_lookup[disk_name] = lun_id
+            else:
+                target_config = self.config.config['targets'][self.target_iqn]
+                used_lun_ids = self._get_lun_ids(target_config['clients'])
+                for disk_name in image_list:
+                    disk_lun_id = target_config['disks'][disk_name]['lun_id']
+                    if disk_lun_id not in used_lun_ids:
+                        self.lun_lookup[disk_name] = disk_lun_id
+                self.requested_images = image_list
+
+    def _get_lun_ids(self, clients_config):
+        lun_ids = []
+        for client_config in clients_config.values():
+            for lun_config in client_config['luns'].values():
+                lun_ids.append(lun_config['lun_id'])
+        return lun_ids
 
     def setup_luns(self, disks_config):
         """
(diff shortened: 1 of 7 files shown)


rbd-target-api.py
```diff
@@ -2493,7 +2497,7 @@ def get_settings():
         'config': {
             'minimum_gateways': settings.config.minimum_gateways
         },
-        'api_version': 2
+        'api_version': 3
```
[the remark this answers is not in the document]
ricardoasmarques: I've changed the order of the commits so we can start with `api_version` 1.

ceph_iscsi_config/lun.py
```diff
@@ -433,15 +433,16 @@ def _get_next_lun_id(self, target_disks):
...
         disk_metadata['owner'] = owner
         self.config.update_item("disks", disk, disk_metadata)
 
         target_disk_config = target_config['disks'].get(disk)
         if not target_disk_config:
-            lun_id = self._get_next_lun_id(target_config['disks'])
+            if lun_id is None:
+                lun_id = self._get_next_lun_id(target_config['disks'])
             target_config['disks'][disk] = {
                 'lun_id': lun_id
```
mikechristie: Did you also want the lun_id to be used as the lun_id when mapped to a client, or did you also want to control that value at the client level?

You can do:

/iscsi-target...h-iscsi/disks> add rbd/testlun 100

but then if you map it to a client:

/iscsi-target...com.ceph:ini1> disk add rbd/testlun

the LUN is just the first available number:

  o- lun 0 ............................ [rbd/test1(1G), Owner: ans6

ricardoasmarques: @mikechristie looking into how `lrbd` works I see that the mapped lun also uses the specified lun_id:

Screenshot from 2019-10-21 12-24-16

Thanks for the hint, I'll adapt this PR.

ricardoasmarques: @mikechristie this should be fixed now.
Clients and groups will now read the `lun_id` that is already defined at target disk level (instead of generating a new one).

ricardoasmarques:
I found an issue in this PR so please **don't merge it yet**.

Assuming we have the following configuration **before** this PR:

```
{
    ...
    "targets": {
        "iqn.2001-07.com.ceph:1571910286153": {
            ...
            "clients": {
                "iqn.2001-07.com.ceph:1571910286153-c1": {
                    ...
                    "luns": {
                        "rbd/img2": {
                            "lun_id": 0
                        }
                    }
                }
            },
            ...
            "disks": {
                "rbd/img1": {
                    "lun_id": 0
                },
                "rbd/img2": {
                    "lun_id": 1
                }
            },
            ...
}
```

Then, **after** this PR, if we try to add `rbd/img1` to `iqn.2001-07.com.ceph:1571910286153-c1` client, we will get an error:

```
/iscsi-target...1910286153-c1> disk add rbd/img1
disk add for 'rbd/img1' against iqn.2001-07.com.ceph:1571910286153-c1 failed
client masking update failed on node1.ceph.local. Client update failed - rbd/img1 is missing from the tpg - unable to map
```

This happens because we will try to map the client lun using the `lun_id 0` that is specified at target-level, but that `lun_id 0` is already in use at the client-level.

**Possible solution:**
When mapping a lun to a client, `ceph-iscsi` should try to use the `lun_id` defined at target-level, but if that `lun_id` is already in use at client-level, then `ceph-iscsi` will find a different `lun_id` that can be used for that client

ricardoasmarques:
> I found an issue in this PR so please **don't merge it yet**.
> ...

It should be fixed now.

mikechristie:
> **Possible solution:**
> When mapping a lun to a client, `ceph-iscsi` should try to use the `lun_id` defined at target-level, but if that `lun_id` is already in use at client-level, then `ceph-iscsi` will find a different `lun_id` that can be used for that client

I think this is a sane solution. I think it would be a user mistake if they cared about the LUN value and they set it up so you would end up in a situation like this.

ceph_iscsi_config/client.py
```diff
@@ -114,6 +98,34 @@ def __init__(self, logger, client_iqn, image_list, username, password, mutual_us
...
+                for disk_item in image_list:
+                    disk_name = disk_item[0]
+                    lun_id = disk_item[1].get('lun_id')
+                    self.requested_images.append(disk_name)
+                    self.lun_lookup[disk_name] = lun_id
+            else:
+                for disk_name in image_list:
+                    target_config = self.config.config['targets'][self.target_iqn]
+                    disk_lun_id = target_config['disks'][disk_name]['lun_id']
+                    used_lun_ids = self._get_lun_ids(target_config['clients'])
```
dillaman: Nit: can `used_lun_ids` and `target_config` be moved outside the loop?

ricardoasmarques: Fixed.

PR #162, opened 2019-12-11
Fix config upgrade from 2 to 3

mikechristie:
If there is no target defined we will crash when upgrading from to 2 to
3. This modifies the version 3 upgrade so we do not assume there will
always be a target defined.


mikechristie:
The ansible people reported it:

https://github.com/ceph/ceph-ansible/pull/4825#event-2874341961

PR #165, opened 2019-12-25
Modify the configuration file name

limin0801:
The ceph documentation indicates that the name of the configuration file created on the iSCSI gateway node is iscsi-gateway.cfg, but the name in the code is gateway.conf, which will cause the service to fail to start properly.

### DIFF
--- a/ceph_iscsi_config/common.py
+++ b/ceph_iscsi_config/common.py
@@ -75,7 +75,7 @@ class Config(object):
 
     lock_time_limit = 30
 
-    def __init__(self, logger, cfg_name='gateway.conf', pool=None):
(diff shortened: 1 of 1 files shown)


mikechristie:
I think you might be confusing 2 config files.

gateway.conf is the name of the config object which stores the gateway definitions (the targets, LUNs, clients, etc). It is stored in the rbd pool by default.

iscsi-gateway.cfg is a file which stores settings. The users creates this file in /etc/ceph on each gateway and it can be used to override the default values for things like IO size limits, timeouts, etc  (the settings you see in at the bottom of ceph_iscsi_config/gateway_settings.py file).

PR #170, opened 2020-01-08
Misc fixes

mikechristie:
This fixes some misc bugs:

1. If no targets are mapped to a gateway rbd-target-gw crashes.
2. rbd-target-api crashes in python3 setups if there are blacklist entries.
3. gateway deletion does not fully cleanup the kernel on the local node.

### DIFF
--- a/ceph_iscsi_config/gateway.py
+++ b/ceph_iscsi_config/gateway.py
@@ -37,23 +37,25 @@ class CephiSCSIGateway(object):
                          "{}".format(blacklisted_ip))
 
         conf = settings.config
-        result = subprocess.check_output("ceph -n {client_name} --conf {cephconf} "
-                                         "osd blacklist rm {blacklisted_ip}".
-                                         format(blacklisted_ip=blacklisted_ip,
-                                                client_name=conf.cluster_client_name,
-                                                cephconf=conf.cephconf),
-                                         stderr=subprocess.STDOUT, shell=True)
-        if ("un-blacklisting" in result) or ("isn't blacklisted" in result):
-            self.logger.info("Successfully removed blacklist entry")
-            return True
-        else:
-            self.logger.critical("blacklist removal failed. Run"
+        try:
+            subprocess.check_output("ceph -n {client_name} --conf {cephconf} "
+                                    "osd blacklist rm {blacklisted_ip}".
+                                    format(blacklisted_ip=blacklisted_ip,
+                                           client_name=conf.cluster_client_name,
+                                           cephconf=conf.cephconf),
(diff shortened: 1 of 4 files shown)


ceph_iscsi_config/gateway.py
```diff
@@ -37,24 +37,26 @@ def ceph_rm_blacklist(self, blacklisted_ip):
...
-            return True
-        else:
-            self.logger.critical("blacklist removal failed. Run"
+        try:
+            subprocess.check_output("ceph -n {client_name} --conf {cephconf} "
+                                    "osd blacklist rm {blacklisted_ip}".
+                                    format(blacklisted_ip=blacklisted_ip,
+                                           client_name=conf.cluster_client_name,
+                                           cephconf=conf.cephconf),
+                                    stderr=subprocess.STDOUT, shell=True)
```
dillaman: Don't you need `check=True` to have it raise an exception? Does the ceph CLI use a non-zero return code for the "isn't blacklisted" case? Could also probably keep the old code but add a `results = results.decode('utf-8')` to ensure bytes are converted to a string.

mikechristie: The "isn't blacklisted" case returns 0.

You don't need to pass in check=True when using check_output:

https://docs.python.org/3/library/subprocess.html

says:

> If the return code was non-zero it raises a CalledProcessError. The CalledProcessError object will have the return code in the returncode attribute and any output in the output attribute.
> 
> This is equivalent to:
> 
> run(..., check=True, stdout=PIPE).stdout

I tested the failure case for both python 3 and 2 and you get an exception.

Yes, we could do a "results.decode('utf-8')" instead.

Let me know if you want me to go the decode() route since you know the "ceph osd" command output a lot better than me.

dillaman: Does that mean that it was always (silently) throwing an exception if `ceph` failed? If you have tested it on the happy path of blacklist successfully removed and blacklist IP didn't exist, I'm happy w/ it.

mikechristie: Yes. It was silently throwing an exception. The thing is its kind of hard to make it return failure under normal circumstances, so we probably never saw it. For probably the common/easy failure case where the cluster is not reachable the command just hangs/waits for at least 5 minutes (I gave up waiting for a failure after 5 min).

I had to test it by disabling the network when doing the "ceph osd blacklist rm" command and adding the arg "--connect-timeout 5", so the command would return failures within a normal time. The "ceph osd blacklist rm" command would then return failure and check_output would raise an exception.

PR #173, opened 2020-02-20
iscsi: Attempt to pull SSL crt and key's from mon config-key store

matthewoliver:
I'm looking at attempting to add iscsi to cephadm, but frankly having
to place the crt and key inside the container makes things a little
harder. While working on adding RGW SSL support to cephadm sage et al,
suggested we should use the mon config-key store the certs[0][1].

So thought why not fo the same here. This will mean when setting up
service in cephadm, an op can either place the certs:

  ceph config-key put 'iscsi/<client-name>/iscsi-gateway.crt -i <file>
  ceph config-key put 'iscsi/<client-name>/iscsi-gateway.key -i <file>

Or if given the files via cephadm we can place them on their behalf.

[0]- https://github.com/ceph/ceph/pull/32951
[1]- https://github.com/ceph/ceph/pull/33287

### DIFF
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -10,6 +10,7 @@ from logging.handlers import RotatingFileHandler
 import ssl
 import operator
 import OpenSSL
+import tempfile
 import threading
 import time
 import inspect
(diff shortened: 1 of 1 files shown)


matthewoliver:
This is still a WIP, just wanted to post it somewhere. Still needs debugging, testing, docs and well tests :P 

But thought better it was done from the beginning in the open, so here it is :)

If someone could add the DNM tag then that'll be great.

rbd-target-api.py
```diff
@@ -2816,13 +2817,46 @@ def run(self):
...
+    temp_files = []
+    with rados.Rados(conffile=settings.config.cephconf,
+                     name=client_name) as cluster:
+        cmd = {"prefix": "mon config-key get",
+               "key": "iscsi/{}/iscsi-gateway.crt".format(client_name)}
+        ret, crt_data, outs = cluster.mon_command(json.dumps(cmd), b'')
+        if not ret:
+            return temp_files
+
+        cmd["key"] = "iscsi/{}/iscsi-gateway.key".format(client_name)
```
dillaman: I would need to double-check from the `cephadm`-side, but for security reasons, is the private key locked down so only the iSCSI user can read them?

rbd-target-api.py
```diff
@@ -2816,13 +2817,46 @@ def run(self):
...
+        if not ret:
+            return temp_files
+    for data in crt_data, key_data:
+        # NOTE: Annoyingly SSLContext.load_cert_chain can only take
+        # paths to files and not file like objects.. yet. So we need to
+        # create tempfiles for the SSL context to read. Once
+        # https://bugs.python.org/issue16487 is resolved, we should be able
+        # to simply use file-like objects and makes this much nicer.
+
+        tmp_f = tempfile.TemporaryFile()
```
dillaman: Nit: you would need to keep this `tmp_f` in-scope so that it doesn't get garbage collected. Otherwise there is a potential race of the temp files being deleted prior to use.

rbd-target-api.py
matthewoliver: I don't think it's limited to the iscsi user unfortunately. Though you would need a keyring with I guess mon read caps.. I haven't tested this. So you'd at least need to be authenticated.

But a very interesting point

rbd-target-api.py
matthewoliver: Ahh yeah, very good spot. I originally inlined this in the `get_ssl_context` but started getting a bit big so pulled it out, yes this could very much be an issue!

rbd-target-api.py
jecluis: using mon caps you could limit access to this prefix for just the iscsi user though. We use similar things for the mgr, for instance, although we achieve that in particular by hardcoding those caps via the mgr profile.

In https://docs.ceph.com/docs/master/rados/operations/user-management/#authorization-capabilities you will find a few examples of this, but I would imagine you'd want to limit access to that key via

> mon 'allow command "config-key get" with key prefix iscsi/$client_name/iscsi-gateway.key,allow command "config-key set" with key prefix iscsi/$client_name/iscsi-gateway.key"

If you want to see a lot more examples of how this could be achieved, you can take a look at `qa/workunits/mon/test_config_key_caps.sh`.

sebastian-philipp:
ping @ricardoasmarques  + @matthewoliver  anything I can do to unstuck this?

matthewoliver:
I've finally had a chance to test, and with the latest force push it is now getting the SSL and key out of the config-key store.

So @jschmid1 you might need to update what your testing in ci. had a few bool checks the wrong way. And the mon_command slightly wrong.

So from that point of view the code is working. I mean with the security questions aside.

dillaman:
(I left a comment for addressing the `config-key get` caps in the other PRs)

sebastian-philipp:
ping @matthewoliver ?

PR #175, opened 2020-03-12
Fix problem when api_host is config.

zhangdaolong:
gwcli will always hold,  when api_host is set but localhost.


gwcli/storage.py
```diff
@@ -330,8 +331,9 @@ def create_disk(self, pool=None, image=None, size=None, count=1,
                 else:
                     disk_key = "{}/{}".format(pool, image)
 
-                disk_api = ('{}://localhost:{}/api/disk/'
+                disk_api = ('{}://{]}:{}/api/disk/'
```
dillaman: Nit: typo

zhangdaolong: thank you,i will fix.

zhangdaolong:
@dillaman i closed this PR,and repoen it.

PR #176, opened 2020-03-13
Fix problem when api_host set in config.

zhangdaolong:
when api_host is config ,command gwcli may be hold.


rbd-target-api.py
```diff
@@ -78,7 +78,7 @@ def decorated(*args, **kwargs):
         # First check that the source of the request is actually valid
         gw_names = [gw for gw in config.config['gateways']
                     if isinstance(config.config['gateways'][gw], dict)]
-        gw_names.append('localhost')
+        gw_names.append(settings.config.api_host)
```
dillaman: The API needs to run certain operations locally before remotely, so these shouldn't change.

zhangdaolong:
i reopen it .

PR #177, opened 2020-03-18
Fix resolve_ip_addresses formatting

mikechristie:
This has resolve_ip_addresses return addresses in the format expected by
the caller. Specifically, requires_restricted_auth expects that the
interface scope and leading "::ffff:" are dropped.

Note:
I am not 100% sure when we even see this. We have had a couple users hit the issue though. One known case is when the user mis-sets the IP in the /etc/hosts file. In this case if you just added an extra dot:

192.168.1..100 somehost.local

then getaddrinfo will return both the link local ipv6 addr and a corrected ipv4 addr. If the address is correct though, then we get the ipv4 address only.

In other cases, we have not seen any misconfigs, so I am guessing it might be differences in python versions, or expected behavior in certain configs. However, I am not sure what those are right now.

### DIFF
--- a/ceph_iscsi_config/utils.py
+++ b/ceph_iscsi_config/utils.py
@@ -86,7 +86,11 @@ def resolve_ip_addresses(addr):
         infos = socket.getaddrinfo(addr, 0)
         for info in infos:
             if info[0] in families:
-                addrs.add(info[4][0])
+                # remove interface scope suffix and IPv4-over-IPv6 prefix
+                addr = info[4][0]
+                addr = addr.rsplit('%', 1)[0]
+                addr = addr.split('::ffff:', 1)[-1]
+                addrs.add(addr)
     except Exception:
         pass
 


ceph_iscsi_config/utils.py
```diff
@@ -86,7 +86,11 @@ def resolve_ip_addresses(addr):
         infos = socket.getaddrinfo(addr, 0)
         for info in infos:
             if info[0] in families:
-                addrs.add(info[4][0])
+                # remove interface scope suffix and IPv4-over-IPv6 prefix
+                addr = info[4][0]
+                addr = addr.rsplit('%', 1)[0]
```
dillaman: Nit: should we just skip over any link-local addresses? i.e. if it contains `%`, just go to the next one?

ceph_iscsi_config/utils.py
```diff
@@ -86,7 +86,11 @@ def resolve_ip_addresses(addr):
...
+                addr = addr.split('::ffff:', 1)[-1]
```
dillaman: I think we want to skip any that start w/ '::ffff' (not just contains '::')

ceph_iscsi_config/utils.py
mikechristie: The bug/issue here was that requires_restricted_auth gets the link local address sometimes (I have still have only been able to replicate it by adding a mis-set address, and cannot figure out why other users see it) in the request.remote_addr value. That function would chop off the zone/interface info after the %, so above I chopped it to match what the caller was expecting. If I do not chop the interface and return the value then we will fail to match in requires_restricted_auth:

```
        if remote_addr not in gw_ips:
            return jsonify(message="API access not available to "
                                   "{}".format(remote_addr)), 403
```

An alt fix would be to just remove the line in requires_restricted_auth:

`remote_addr = request.remote_addr.rsplit('%', 1)[0]
`
but I was thinking there might be future resolve_ip_addreses users that want the address. I am really not sure though.

I do not know ipv6 well, but I noticed you added the requires_restricted_auth code. The only time I have seen the zone info is when it has the interface info for link local addresses, but it sounds like you can get the zone info for non link local addresses.

Did you by any chance see the interface info after the "%" and add that code in requires_restricted_auth for the for link local case?

ceph_iscsi_config/utils.py
mikechristie: I have not actually ever seen it returned. I only did that to match what requires_restricted_auth was doing:

requires_restricted_auth gets "::ffff:192.168.1.100" in request.remote_addr. That function drops the ::ffff, and then socket.getaddrinfo always returns "192.168.1.100".

I just added it above just in case there was a case getaddrinfo could return  "::ffff:192.168.1.100" instead of "192.168.1.100" and we just have not seen it.

ceph_iscsi_config/utils.py
dillaman: I've only seen scopes on link-local addresses -- and those are added to each interface where IPv6 is enabled (even if the same interface has a real IPv6 address). I'm fine dropping support for link-local (i.e. strip them out everywhere) if that's doable.

ceph_iscsi_config/utils.py
dillaman: Gotcha. Yeah, those IPv4-over-IPv6 addresses technically need to start with '::ffff:' and not just a random `::ffff` anywhere in the address (i.e. since `1234::ffff:abcd` isn't an IPv4-over-IPv6 address)

mikechristie: Ah I see what you are saying. I misunderstood how split() works. I can fix that up.

I was thinking why are you saying the split() use is wrong, because I thought you wrote it and I am just copying you :)

dillaman: Lol -- nope, I was wrong and it would inadvertently break any IPv6 addresses with an embedded `::ffff`

mikechristie:
Dropping this for now, because it looks like it might have been another bad setup.

Will open a new PR to handle all the issues that have come up like the ::ffff parsing and general link local handling.

PR #181, opened 2020-03-21
Fix scan rbd image stuck

zhouyh-orcadt:
client.py :
config submission may fail, there should be a return value judgment.

storage.py ：
when the rbd image is full, it cannot operate because of the write lock.
“with rbd.Image (ioctx, image) as rbd_image:” stuck,
it can be adjusted to read-only mode to read image information.

### DIFF
--- a/ceph_iscsi_config/client.py
+++ b/ceph_iscsi_config/client.py
@@ -665,6 +665,7 @@ class GWClient(GWObject):
                         config_object.commit()
                         if config_object.error:
                             self.error = True
+                            self.error_msg = config_object.error_msg
 
         elif rqst_type == 'reconfigure':
             self.define_client()
@@ -689,6 +690,7 @@ class GWClient(GWObject):
                         config_object.commit()
                         if config_object.error:
                             self.error = True
+                            self.error_msg = config_object.error_msg
 
             else:
                 # desired state is absent, but the client does not exist
--- a/ceph_iscsi_config/client.py
+++ b/ceph_iscsi_config/client.py
@@ -663,6 +663,8 @@ class GWClient(GWObject):
 
                         # persist the config update
                         config_object.commit()
+                        if config_object.error:
+                            self.error = True
 
         elif rqst_type == 'reconfigure':
             self.define_client()
@@ -685,6 +687,8 @@ class GWClient(GWObject):
                         target_config['clients'].pop(self.iqn)
                         config_object.update_item("targets", self.target_iqn, target_config)
                         config_object.commit()
+                        if config_object.error:
+                            self.error = True
 
             else:
                 # desired state is absent, but the client does not exist
--- a/ceph_iscsi_config/common.py
+++ b/ceph_iscsi_config/common.py
@@ -482,13 +482,12 @@ class Config(object):
 
     def _seed_rbd_config(self):
 
-        ioctx = self._open_ioctx()
-
         self.lock()
         if self.error:
             return
 
         # if the config object is empty, seed it - if not just leave as is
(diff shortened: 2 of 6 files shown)


gwcli/storage.py
```diff
@@ -76,7 +76,7 @@ def _get_disk_meta(self, cluster_ioctx, disk_meta):
                 disk_meta[rbd_name] = {}
                 with cluster_ioctx.open_ioctx(pool) as ioctx:
                     try:
-                        with rbd.Image(ioctx, image) as rbd_image:
+                        with rbd.Image(ioctx, image, read_only=True) as rbd_image:
```
mikechristie: I can't seem to replicate the bug that this was fixing. Can you tell me how you are confirming the disk is full?

ceph_iscsi_config/client.py
```diff
@@ -663,6 +663,8 @@ def manage(self, rqst_type, committer=None):
 
                         # persist the config update
                         config_object.commit()
+                        if config_object.error:
+                            self.error = True
```
mikechristie: You should set the error_msg here and below.

zhouyh-orcadt: Yes, you are right, i will increase it.

ceph_iscsi_config/common.py
```diff
@@ -482,13 +482,12 @@ def unlock(self):
 
     def _seed_rbd_config(self):
```
mikechristie, line 484: The open/lock related changes are not related to your original issue right? They fix things like leaks.

Could you make this into a proper patchset? If this is one of your first patchsets to an upstream project here are some hints/guidelines that I use:

1. When upstream merges your PR into master, we don't want to see your git history. For example we do not want to see this:

Merge pull request #1 from zhouyh-orcadt/fix_scan_image_stuck
@zhouyh-orcadt
zhouyh-orcadt committed on Mar 20

and we want to only see the final patches. We do not want to see version 1 of a patch then a fix to that.

Instead of adding patch on top of patches you can/should also do something like a force push.

So let's say you fix some issue and make a PR. You then get a review comment asking you to add some extra code to the patch. In your local git tree branch you can revert your original patch, fix up the code, commit it. Then do

git push --force your_branch master

This will basically overwrite the old branch with the new commit and in the git history we only see the final patch which is all we care about in the main tree. The nice thing is that this will also automatically update the PR with the new code, so you do not have to do open a new PR.

2. Try to do one cleanup/fix in each patch. We currently have a mix and match of stuff going on in each patch. Later if we have a regression and need to revert, or need to cherry pick a specific patch, we can't because the changes are spread over multiple patches.

3. Add a description in the git commit, so later when someone does a "git log" they know what the patch does.

So for example for 2 and 3, you would have 3 patches.

patch 1 - fix the close/lock issues. In the patch description you could write something like:

This patch fixes an issue where the common.py is cleaning the ioctx open if a lock call fails.

patch 2- fix the error_msg issue. In the patch description you could write something like:

This patch fixes an issue where we are not checking for config commit errors and returning a error_msg in a couple places.

patch 3 - fix the RO issue. In the patch description you could write what you wrote here:

https://github.com/ceph/ceph-iscsi/pull/181#discussion_r396291475

It's a really good desrciption of the problem and how your patch solves it.

mikechristie: Also, thanks for the fixing all these issues.

Let me know if you do not have time to fix up the patchset.

I can do it manually, and you will get the credit in the git commit logs, but I think due to a bug in GH it will not could in your GH stats if you care about them for work or personal reasons.

PR #183, opened 2020-04-02
Fix problem when api_host set in config.

zhangdaolong:
when api_host is config ,command gwcli may be hold.


gwcli.8
```diff
@@ -31,6 +31,8 @@ Typically, the following options are regarded as site specific;
 .PP
 \fBapi_password = <password>\fR
 .PP
+\fBapi_host = <ip_address>\fR
```
mikechristie, line 34: How about:

<ip_address or hostname>

rbd-target-api.py
```diff
@@ -2621,6 +2621,8 @@ def target_ready(gateway_list):
                     "summary": ''}
 
     for gw in gateway_list:
+        if gw == "localhost"
```
mikechristie, line 2624: Need a ":" at the end of the line.

`if gw == "localhost":`

rbd-target-api.py
```diff
@@ -2673,6 +2675,9 @@ def call_api(gateway_list, endpoint, element, http_method='put', api_vars=None):
     logger.debug("gateway update order is {}".format(','.join(gateway_list)))
 
     for gw in gateway_list:
+        if gw == "localhost"
```
mikechristie, line 2678: Same as above.

ceph_iscsi_config/gateway_setting.py
```diff
@@ -171,7 +171,9 @@ def normalize(self, raw_val):
     "pool": StrSetting("pool", "rbd"),
     "cluster_client_name": StrSetting("cluster_client_name", "client.admin"),
     "time_out": IntSetting("time_out", 1, 600, 30),
-    "api_host": StrSetting("api_host", "::"),
+    # if the api_host has a colon, you must wrap it with square brackets.
+    # i.e:[::]
+    "api_host": StrSetting("api_host", "localhost"),
     "api_port": IntSetting("api_port", 1, 65535, 5000),
     "api_secure": BoolSetting("api_secure", True),
     "api_ssl_verify": BoolSetting("api_ssl_verify", False),
```
mikechristie, line 179: You will need to add api_host to exclude_from_hash in ceph_iscsi_config/setting.py, because that value can be different on each node.

rbd-target-api.py
mikechristie: One question just to make sure I understood the localhost check. Is it for cases like rbd-target-api:clientauth():

`    gateways.insert(0, 'localhost')
`
The check you are adding allows callers to just use localhost instead of doing 

gateways.insert(0, settings.config.api_host)

right?

zhangdaolong: yes ,that's what I thought.

gwcli.8
zhangdaolong: good idea

ceph_iscsi_config/gateway_setting.py
zhangdaolong: ok i add it

zhangdaolong:
@mikechristie 
I have revised it, please take a review, thank you

ceph_iscsi_config/gateway_setting.py
```diff
@@ -171,7 +171,9 @@ def normalize(self, raw_val):
     "pool": StrSetting("pool", "rbd"),
     "cluster_client_name": StrSetting("cluster_client_name", "client.admin"),
     "time_out": IntSetting("time_out", 1, 600, 30),
-    "api_host": StrSetting("api_host", "::"),
+    # if the api_host has a colon, you must wrap it with square brackets.
+    # i.e:[::]
+    "api_host": StrSetting("api_host", "localhost"),
```
mikechristie, line 176: Hey

Sorry for the delay on this, but I just can't get it to work and I'm not sure if its my setup or the patch.

1. If I use the defaults set in the patch, and I do not override anything, then I get "Connection refused" errors when trying to do any command to the the daemon.

Does using the default values work for you? If so, when yo do:

ping localhost

do you see localhost resolving to ::1 or 127.0.1.

2. If you use "[::]" like documented in the code comments above, then rbd-target-api crashes.

3. If I use "::"  as the default above or in iscsi-gateway.cfg for the api_host value it works ok for me then.

zhangdaolong: > Hey
> 
> Sorry for the delay on this, but I just can't get it to work and I'm not sure if its my setup or the patch.
> 
>     1. If I use the defaults set in the patch, and I do not override anything, then I get "Connection refused" errors when trying to do any command to the the daemon.
> 
> 
> Does using the default values work for you? If so, when yo do:
> 
> ping localhost
> 
> do you see localhost resolving to ::1 or 127.0.1.

It is better to use ip such as the host name, because there is no mapping configured for localhost。
If localhost host mapping is configured on /etc/hosts, there should be no problem。
ie:/etc/hosts
[root@ceph ~]# cat /etc/hosts
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6


[root@ceph ~]# ping localhost
PING localhost (127.0.0.1) 56(84) bytes of data.
64 bytes from localhost (127.0.0.1): icmp_seq=1 ttl=64 time=0.014 ms
^C
--- localhost ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 0.014/0.014/0.014/0.000 ms
[root@ceph ~]# 



> 
>     1. If you use "[::]" like documented in the code comments above, then rbd-target-api crashes.
> 

>     2. If I use "::"  as the default above or in iscsi-gateway.cfg for the api_host value it works ok for me then.

A does not support IPs that contain [and] characters，so i change it
   1     # Start the API server. threaded is enabled to prevent deadlocks when one
2893     # request makes further api requests                                                                                                                                                                      
   1     app.run(host=settings.config.api_host.strip('[').strip(']'),
......

I test it ,it works for me .

PR #185, opened 2020-04-30
Fix rbd.image scan stuck

zhouyh-orcadt:
Signed-off-by: yaohui.zhou <yaohui.zhou@orcadt.com>

update 3 patches:
	patch 1 - fix the close/lock issues. This patch fixes an issue where the common.py is cleaning the ioctx open if a lock call fails.
	patch 2 - fix the error_msg issue. This patch fixes an issue where we are not checking for config commit errors and returning a error_msg in a couple places.
	patch 3 - fix the rbd.image scan stuck issue. This patch fixes an issue where the pool is full, the code "with rbd.Image (ioctx, image) as rbd_image:" in gwcli will get stuck.

### DIFF
--- a/ceph_iscsi_config/common.py
+++ b/ceph_iscsi_config/common.py
@@ -482,13 +482,12 @@ class Config(object):
 
     def _seed_rbd_config(self):
 
-        ioctx = self._open_ioctx()
-
         self.lock()
         if self.error:
             return
 
         # if the config object is empty, seed it - if not just leave as is
+        ioctx = self._open_ioctx()
         cfg_data = self._read_config_object(ioctx)
         if not cfg_data:
             self.logger.debug("_seed_rbd_config found empty config object")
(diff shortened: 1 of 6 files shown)


ceph_iscsi_config/common.py
```diff
@@ -482,13 +482,12 @@ def unlock(self):
 
     def _seed_rbd_config(self):
 
-        ioctx = self._open_ioctx()
```
dillaman, line 485: Nit: perhaps just use a `with self._open_ioctx() as ioctx:` block ....

ceph_iscsi_config/common.py
```diff
@@ -591,15 +591,14 @@ def set_item(self, cfg_type, element_name, element_value):
 
     def _commit_rbd(self, post_action):
 
-        ioctx = self._open_ioctx()
```
dillaman, line 594: Nit: same comment here

PR #186, opened 2020-05-07
Add a config option to log to stderr

matthewoliver:
Most the other ceph components have an option to get them to log to
stderr. And this is used when containerising to help gather logs.

This patch adds a config options `log_to_stderr`, `log_to_stderr_prefix`
and `log_to_file` to do the same for ceph-iscsi. These are matching the
other ceph daemon's:

  --default-log-to-stderr=true --default-log-stderr-prefix="debug "
  --default-log-to-file=false

### DIFF
--- a/ceph_iscsi_config/gateway_setting.py
+++ b/ceph_iscsi_config/gateway_setting.py
@@ -190,6 +190,9 @@ SYS_SETTINGS = {
     "prometheus_host": StrSetting("prometheus_host", "::"),
     "logger_level": IntSetting("logger_level", logging.DEBUG, logging.CRITICAL,
                                logging.DEBUG),
+    "log_to_stderr": BoolSetting("log_to_stderr", False),
+    "log_to_stderr_prefix": StrSetting("log_to_stderr_prefix", ""),
+    "log_to_file": BoolSetting("log_to_file", True),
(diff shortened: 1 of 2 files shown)


dillaman:
(minus the tox failure)

matthewoliver:
hmm, I don't get a tox error locally, and it seems to be complaining actually failing to call the flake8 binary.

What's the magic to get it to rerun the test?

dillaman:
The error was: `./rbd-target-api.py:2968:87: E999 SyntaxError: invalid syntax`

You can re-tick `tox` by addding the following comment: `jenkins tox`

matthewoliver:
Oh yes, sorry apparently I fail at reading logs :P
OK, so that's because flake is using py2. OK, so need to use a py2 compatible version.

matthewoliver:
jenkins tox

matthewoliver:
Hrmm, I get these same errors when I run py2 version of flake8, even on master. But if I just run tox locally it's fine (as I assume it's running py3).

Ceph has gone py3 only now that octopus is out, should we just drop py3? Bigger question though is why is this happening now?

dillaman:
Yeah -- the errors have changed now and they are picking up bugs from a year ago. Very strange (and now unrelated to this PR).

dillaman:
jenkins tox

PR #187, opened 2020-05-07
don't store to config if create_target fails

wuxingyi:
Signed-off-by: wuxingyi <wuxingyi2015@outlook.com>

### DIFF
--- a/ceph_iscsi_config/target.py
+++ b/ceph_iscsi_config/target.py
@@ -681,6 +681,9 @@ class GWTarget(GWObject):
             else:
                 # create the target
                 self.create_target()
+                # if error happens, we should never store this target to config
+                if self.error:
+                    return
                 seed_target = {
                     'disks': {},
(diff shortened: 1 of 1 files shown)


ceph_iscsi_config/target.py
```diff
@@ -681,33 +681,37 @@ def manage(self, mode):
...
-                                                 discovery_auth_config['password'],
-                                                 discovery_auth_config[
-                                                     'password_encryption_enabled'],
-                                                 discovery_auth_config['mutual_username'],
-                                                 discovery_auth_config['mutual_password'],
-                                                 discovery_auth_config[
-                                                     'mutual_password_encryption_enabled'])
+                # if error happens, we should never store this target to config
+                if self.error:
+                    self.logger.error("Failed to create target")
```
dillaman, line 686: Nit: just `return` on error and then you can avoid the `else` branch and associated indentation

Issue #193, opened 2020-06-08
gwcli fails to create gateway on ubuntu 20.04


hemna:
When running gwcli to create a gateway, gwcli dumps out with

```
/iscsi-target...ndev/gateways> create devstack-20-2 192.168.122.174 skipchecks=true
CMD: ../gateways/ create devstack-20-2 ['192.168.122.174'] nosync=False skipchecks=true
OS version/package checks have been bypassed
Adding gateway, sync'ing 0 disk(s) and 0 client(s)
Gateway creation successful
Adding gw to UI
Traceback (most recent call last):
  File "/usr/bin/gwcli", line 194, in <module>
    main()
  File "/usr/bin/gwcli", line 125, in main
    shell.run_interactive()
  File "/usr/lib/python3/dist-packages/configshell_fb/shell.py", line 900, in run_interactive
    self._cli_loop()
  File "/usr/lib/python3/dist-packages/configshell_fb/shell.py", line 729, in _cli_loop
    self.run_cmdline(cmdline)
  File "/usr/lib/python3/dist-packages/configshell_fb/shell.py", line 843, in run_cmdline
    self._execute_command(path, command, pparams, kparams)
  File "/usr/lib/python3/dist-packages/configshell_fb/shell.py", line 818, in _execute_command
    result = target.execute_command(command, pparams, kparams)
  File "/usr/lib/python3/dist-packages/configshell_fb/node.py", line 1406, in execute_command
    return method(*pparams, **kparams)
  File "/usr/lib/python3/dist-packages/gwcli/gateway.py", line 868, in ui_command_create
    gateway_hostname = api.response.json()['data']
  File "/usr/lib/python3/dist-packages/requests/models.py", line 885, in json
    encoding = guess_json_utf(self.content)
  File "/usr/lib/python3/dist-packages/requests/utils.py", line 871, in guess_json_utf
    nullcount = sample.count(_null)
TypeError: must be str, not bytes
```

targetcli seems happy with the gateway

```
root@devstack-20-2:~# targetcli
targetcli shell version 2.1.51
Copyright 2011-2013 by Datera, Inc and others.
For help on commands, type 'help'.

/> ls
o- / ......................................................................................................................... [...]
  o- backstores .............................................................................................................. [...]
  | o- block .................................................................................................. [Storage Objects: 0]
  | o- fileio ................................................................................................. [Storage Objects: 0]
  | o- pscsi .................................................................................................. [Storage Objects: 0]
[...]

PR #196, opened 2020-06-19
Fix gateway creation crash in python3

xin3liang:
This propose fix issue: https://github.com/ceph/ceph-iscsi/issues/193

When setting api_host to localhost will hit below crash[1] in python3.
Json response should be encoded into utf-xx bytes[2].
And also add an error handling path for hostname api request.

[1]: 
/iscsi-target...ndev/gateways> create devstack-20-2 192.168.122.174
skipchecks=true
CMD: ../gateways/ create devstack-20-2 ['192.168.122.174'] nosync=False
skipchecks=true
OS version/package checks have been bypassed
Adding gateway, sync'ing 0 disk(s) and 0 client(s)
Gateway creation successful
Adding gw to UI
      File "/usr/bin/gwcli", line 194, in <module>
          main()
        File "/usr/bin/gwcli", line 125, in main
          shell.run_interactive()
        File "/usr/lib/python3/dist-packages/configshell_fb/shell.py",
      line 900, in run_interactive
          self._cli_loop()
        File "/usr/lib/python3/dist-packages/configshell_fb/shell.py",
      line 729, in _cli_loop
          self.run_cmdline(cmdline)
        File "/usr/lib/python3/dist-packages/configshell_fb/shell.py",
      line 843, in run_cmdline
          self._execute_command(path, command, pparams, kparams)
        File "/usr/lib/python3/dist-packages/configshell_fb/shell.py",
      line 818, in _execute_command
          result = target.execute_command(command, pparams, kparams)
        File "/usr/lib/python3/dist-packages/configshell_fb/node.py",
      line 1406, in execute_command
          return method(*pparams, **kparams)
        File "/usr/lib/python3/dist-packages/gwcli/gateway.py", line
      868, in ui_command_create
          gateway_hostname = api.response.json()['data']
        File "/usr/lib/python3/dist-packages/requests/models.py", line
      885, in json
          encoding = guess_json_utf(self.content)
        File "/usr/lib/python3/dist-packages/requests/utils.py", line
      871, in guess_json_utf
          nullcount = sample.count(_null)
      TypeError: must be str, not bytes

[2]: https://github.com/psf/requests/issues/4879

Signed-off-by: Xinliang Liu <xinliang.liu@linaro.org>

### DIFF
--- a/gwcli/gateway.py
+++ b/gwcli/gateway.py
@@ -833,6 +833,24 @@ class GatewayGroup(UIGroup):
         if skipchecks == 'true':
             self.logger.warning("OS version/package checks have been bypassed")
 
+        # Check if we can get hostname from
+        # the new gw endpoint
+        new_gw_endpoint = ('{}://{}:{}/'
+                           'api'.format(self.http_mode,
+                                        gateway_name,
+                                        settings.config.api_port))
+        api = APIRequest('{}/sysinfo/hostname'.format(new_gw_endpoint))
+        api.get()
+        if api.response.status_code != 200:
+            msg = response_message(api.response, self.logger)
+            self.logger.error("Get gateway hostname failed : {}\n"
+                              "Please check api_host setting and make sure "
+                              "host {} IP is listening on port {}"
+                              "".format(msg, gateway_name,
+                                        settings.config.api_port))
+            return
+        gateway_hostname = api.response.json()['data']
+
         self.logger.info("Adding gateway, {}".format(sync_text))
 
         gw_api = '{}://{}:{}/api'.format(self.http_mode,
@@ -858,14 +876,6 @@ class GatewayGroup(UIGroup):
         # add to the UI. We have to use the new gateway to ensure what
         # we get back is current (the other gateways will lag until they see
         # epoch xattr change on the config object)
-        new_gw_endpoint = ('{}://{}:{}/'
-                           'api'.format(self.http_mode,
-                                        gateway_name,
-                                        settings.config.api_port))
-
-        api = APIRequest('{}/sysinfo/hostname'.format(new_gw_endpoint))
-        api.get()
-        gateway_hostname = api.response.json()['data']
         config = self.parent.parent.parent._get_config(endpoint=new_gw_endpoint)
         target_config = config['targets'][target_iqn]
         portal_config = target_config['portals'][gateway_hostname]
--- a/gwcli/utils.py
+++ b/gwcli/utils.py
@@ -440,7 +440,7 @@ class APIRequest(object):
                        "{}".format(self.args[0]))
                 self.data = Response()
                 self.data.status_code = 500
-                self.data._content = '{{"message": "{}" }}'.format(msg)
+                self.data._content = '{{"message": "{}" }}'.format(msg).encode('utf-8')
                 return self._get_response
             except Exception:
                 raise GatewayAPIError("Unknown error connecting to "


xin3liang:
Hi, please help to review. @dillaman

gwcli/utils.py
```diff
@@ -440,7 +440,7 @@ def __getattr__(self, name):
                        "{}".format(self.args[0]))
                 self.data = Response()
                 self.data.status_code = 500
-                self.data._content = '{{"message": "{}" }}'.format(msg)
+                self.data._content = '{{"message": "{}" }}'.format(msg).encode('utf-8')
```
dillaman, line 443: Nit: why is this necessary?

gwcli/gateway.py
```diff
@@ -865,6 +865,19 @@ def ui_command_create(self, gateway_name, ip_addresses, nosync=False,
 
         api = APIRequest('{}/sysinfo/hostname'.format(new_gw_endpoint))
         api.get()
+
+        if api.response.status_code != 200:
```
dillaman, line 869: Nit: any chance this `hostname` API get could be moved to before the attempted GW creation? From a quick glance it looks like it's just pulling the hostname so if it could do that first you could avoid the need to attempt to delete the gateway.

gwcli/utils.py
xin3liang: According to [this commit msg](https://github.com/psf/requests/commit/4decc7986e32bb8f3511df3dd0c9b1c1d57453c1) of requests as shown below. [And the type of native string in Python2 is bytes but in Python3 is text](https://portingguide.readthedocs.io/en/latest/strings.html#the-native-string), so we need to convert it into bytes explicitly in Python3.
"
JSON *must* be encoded using UTF-8, UTF-16 or UTF-32 (see the [RFC][1]; detect the encoding based on the fact that JSON always starts with 2 ASCII characters.

[1]: http://tools.ietf.org/html/rfc4627#section-3
"
Related issue: https://github.com/psf/requests/issues/4879

xin3liang: > But Python 3 defaults to `UTF-8` encoding for strings, right [1]?
> 
> [1] https://docs.python.org/3/howto/unicode.html#the-string-type

Yes, you are right @dillaman . Sorry, my previous explanation (especially the Request commit log). which may cause confusion that this issue is related to encoding. But it is not. It is related to obj type.  It's my bad ;-) , just forget my previous explanation now. Let me explain more details about this here.

First,  we need to know that 'str' object type could be converting into 'bytes'  object type via [str.encode() ](https://docs.python.org/3/library/stdtypes.html#str.encode) method.

Then, let's look at the crash error and related code lines of Request below. When it runs line "nullcount = sample.count(_null)", it throws a TypeError exception "TypeError: must be str, not bytes". Look at the Request code lines below, "_null = '\x00'.encode('ascii')" means that '_null' var is a bytes obj type because it calls encode(). The sample var gets its value from response.content, it is 'str' obj type. Then you got "str.count(bytes)" that's why it throws  TypeError exception "TypeError: must be str, not bytes". 
So the solution here is make sample var as a bytes obj type by calling encode() while assigning value to response.content.


Crash error
"...
File "/usr/lib/python3/dist-packages/gwcli/gateway.py", line
868, in ui_command_create
gateway_hostname = api.response.json()['data']
File "/usr/lib/python3/dist-packages/requests/models.py", line
885, in json
encoding = guess_json_utf(self.content)
File "/usr/lib/python3/dist-packages/requests/utils.py", line
871, in guess_json_utf
nullcount = sample.count(_null)
TypeError: must be str, not bytes
"

Request related code lines
"...
[encoding = guess_json_utf(self.content)](https://github.com/psf/requests/blob/master/requests/models.py#L888)
...
[_null = '\x00'.encode('ascii')  # The null is a bytes obj type
...
sample = data[:4]
...
nullcount = sample.count(_null)](https://github.com/psf/requests/blob/master/requests/utils.py#L856-L876)
"

gwcli/gateway.py
xin3liang: Good idea! Let me find out if we can do it. Will update the patch soon. Thanks @dillaman

xin3liang: Done

gwcli/utils.py
dillaman: The original error was because the command failed and yet because the status code wasn't checked for success, it attempted to decode the invalid response as json. I think you'll find that this works just fine now that you added the check for success on the `/sysinfo/hostname` call.

At least it works as-is on my RHEL 8 / Python 3 environment w/ just the status check.

xin3liang: > The original error was because the command failed and yet because the status code wasn't checked for success, it attempted to decode the invalid response as json. I think you'll find that this works just fine now that you added the check for success on the `/sysinfo/hostname` call.
> 
> At least it works as-is on my RHEL 8 / Python 3 environment w/ just the status check.

Yep, I agree that status code checking is indeed necessary here. And normally it will work well.
But, what will happen if we hit the "status_code != 200" path (One way to hit this path is setting api_host = localhost like git msg saying and create a gateway with a non-localhost ip hostname)? I would say it will get the same crash if we don't put encode() at the end of this response.content. Because in the "status_code != 200" path, it also parse json to get message via method response_message(). And this will happen in other api call too not just the gateway creation. 
So we also need to fix this response.content making it as bytes object type to meet Request's requirement and avoid this potential crash.

PR #207, opened 2020-08-04
Support specified gateway config name

zhangdaolong:
Add the option to specify the gateway  configuration file, which can configure multiple gateways config  in the same pool.

Signed-off-by: zhang daolong <zhangdaolong@fiberhome.com>

### DIFF
--- a/ceph_iscsi_config/gateway_setting.py
+++ b/ceph_iscsi_config/gateway_setting.py
@@ -183,6 +183,7 @@ SYS_SETTINGS = {
     "debug": BoolSetting("debug", False),
     "minimum_gateways": IntSetting("minimum_gateways", 1, 9999, 2),
     "ceph_config_dir": StrSetting("ceph_config_dir", '/etc/ceph'),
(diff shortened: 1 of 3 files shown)


ceph_iscsi_config/gateway_setting.py
```diff
@@ -184,6 +184,7 @@ def normalize(self, raw_val):
     "minimum_gateways": IntSetting("minimum_gateways", 1, 9999, 2),
     "ceph_config_dir": StrSetting("ceph_config_dir", '/etc/ceph'),
     "priv_key": StrSetting("priv_key", 'iscsi-gateway.key'),
+    "gateway_conf": StrSetting("gateway_conf", 'iscsi-gateway.conf'),
```
dillaman, line 187: Nit: perhaps keep `priv_key` and `pub_key` next to each other since they are linked.

zhangdaolong: > Nit: perhaps keep `priv_key` and `pub_key` next to each other since they are linked.

Ok,i fix it.thank you .

Issue #208, opened 2020-08-12
api: rados connection error on startup


dsavineau:
When using ceph and ceph-iscsi from chacra repositories, I not able to start rbd-target-api anymore (it was working fine few days ago) on a non containerized environment.

```console
# rpm -qa | egrep '(ceph|rados)'
libcephfs2-16.0.0-4355.g4fa1eaf.el8.x86_64
python3-cephfs-16.0.0-4355.g4fa1eaf.el8.x86_64
python3-ceph-common-16.0.0-4355.g4fa1eaf.el8.x86_64
ceph-common-16.0.0-4355.g4fa1eaf.el8.x86_64
ceph-iscsi-3.4-27.g6193889.el8.noarch
librados2-16.0.0-4355.g4fa1eaf.el8.x86_64
python3-rados-16.0.0-4355.g4fa1eaf.el8.x86_64
python3-ceph-argparse-16.0.0-4355.g4fa1eaf.el8.x86_64
libradosstriper1-16.0.0-4355.g4fa1eaf.el8.x86_64
```

The rbd-target-api log ouput shows that there's a permission denied when trying to connect the ceph cluster even if every requirement are present (ceph config, pool, keyring, etc...).

```console
2020-08-12 17:42:32,786    ERROR [rbd-target-api:2993:<module>()] - Unable to connect to the cluster (keyring missing?) - [errno 13] RADOS permission denied (error connecting to the cluster)
2020-08-12 17:42:32,787 CRITICAL [rbd-target-api:2764:halt()] - Unable to open/read the configuration object
```

I'm using the ceph admin keyring for this and I'm able to run some ceph commands with that keyring.

```console
# ls -hl /etc/ceph/
total 16K
-rw-------. 1 ceph ceph  151 Aug 12 14:43 ceph.client.admin.keyring
-rw-r--r--. 1 ceph ceph  455 Aug 12 14:43 ceph.conf
-rw-------. 1 root root 1012 Aug 12 16:59 iscsi-gateway.cfg
-rw-r--r--. 1 root root   92 Aug 12 09:14 rbdma
```

When trying the cluster connection manually via python (which is what is current implemented in ceph-iscsi [1]) we can see that the error is coming from `rados.pyx`.

```console
>>> import rados
>>> conf = '/etc/ceph/ceph.conf'
>>> cluster = rados.Rados(conffile=conf, name='client.admin')
>>> cluster.connect()
Traceback (most recent call last):
[...]

PR #209, opened 2020-08-12
Fix rados cluster connection

dsavineau:
Without specifying the ceph cluster name via the clustername parameter
on the Rados class then the rbd target api doesn't start anymore with
recent ceph version (16.0.0+).
The cluster name value is already present in the settings via the
cluster_name option.

Closes: #208

Signed-off-by: Dimitri Savineau <dsavinea@redhat.com>

### DIFF
--- a/ceph_iscsi_config/common.py
+++ b/ceph_iscsi_config/common.py
@@ -35,11 +35,13 @@ class CephCluster(object):
         conf = settings.config.cephconf
 
         try:
-            self.cluster = rados.Rados(conffile=conf,
+            self.cluster = rados.Rados(clustername=settings.config.cluster_name,
+                                       conffile=conf,
                                        name=settings.config.cluster_client_name)
         except rados.Error as err:
             self.error = True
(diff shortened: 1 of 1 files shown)


dillaman:
If this is a regression in python-rados, I think it should be fixed there since otherwise lots of old client programs will break. If you use the latest octopus version of python-rados does it function?

dsavineau:
> If you use the latest octopus version of python-rados does it function?

I only tested Octopus 15.2.4 and it was working fine. But something could have been already backported from master to the octopus branch at this point.

dsavineau:
@dillaman Should I open a new rados tracker instead for this ?

dillaman:
> @dillaman Should I open a new rados tracker instead for this ?

Yes, probably a good idea.

dsavineau:
https://tracker.ceph.com/issues/47002

dsavineau:
Since the issue has been fixed on ceph side, I'll close that one

PR #210, opened 2020-08-13
Report tcmu-runner device status (new)

lxbsz:
These patches allow rbd-target-api to check the tcmu-runner ceph status, mark a device as degraded and then report the state. It converts gwcli, but the rbd-target-api REST call can be used for other tools like dashboard.

These patches require tcmu-runner support:
open-iscsi/tcmu-runner#630

Based Mike's old PR#191, I have a small fixing(https://github.com/ceph/ceph-iscsi/commit/578101e005f742a83859e64150656e8a4da02060), It should be fold to the first commit, I just split it into a separate one to make it easier to review. After the reviewing is done I will repush and fold them.


gwcli/storage.py
```diff
@@ -684,6 +686,32 @@ def __init__(self, parent, image_id, image_config, size=None,
...
+                                     settings.config.api_port, self.image_id))
+        self.logger.debug("disk GET status for {}".format(self.image_id))
+        api = APIRequest(disk_api)
+        api.get()
+
+        if api.response.status_code == 200:
+            info = api.response.json()
+            status = info.get("status")
+
+            self.__setattr__('current_owner', 'Unknown')
```
dillaman, line 701: Nit: move this up before the `status_code == 200` check

ceph_iscsi_config/gateway_setting.py
```diff
@@ -204,3 +204,8 @@ def normalize(self, raw_val):
     "qfull_timeout": IntSetting("qfull_timeout", 0, 600, 5),
     "osd_op_timeout": IntSetting("osd_op_timeout", 0, 600, 30),
     "hw_max_sectors": IntSetting("hw_max_sectors", 1, 8192, 1024)}
+
+TCMU_DEV_STATUS_SETTINGS = {
+    "lock_lost_cnt_thresh": IntSetting("lock_lost_cnt_thresh", 1, 1000000, 12),
```
dillaman, line 209:
```suggestion
    "lock_lost_count_threshold": IntSetting("lock_lost_count_threshold", 1, 1000000, 12),
```

ceph_iscsi_config/gateway_setting.py
```diff
@@ -204,3 +204,8 @@ def normalize(self, raw_val):
...
+    "status_check_interval": IntSetting("status_check_interval", 1, 600, 60),
+    "stable_state_reset_cnt": IntSetting("stable_state_reset_cnt", 1, 600, 3)}
```
dillaman, line 211:
```suggestion
    "stable_state_reset_count": IntSetting("stable_state_reset_count", 1, 600, 3)}
```

ceph_iscsi_config/device_status.py
```diff
@@ -0,0 +1,165 @@
...
+            cmd = json.dumps({"prefix": "service status", "format": "json"})
+
+            ret, outb, outs = self.cluster.mon_command(cmd, b'')
+            if ret != 0:
+                self.logger.error("mon command failed {}".format(ret))
+                continue
+
+            svc = json.loads(outb).get('tcmu-runner')
+            if svc is None:
+                continue
```
dillaman, line 125: Nit: can we track in the state that we have no data available from `tcmu-runner`? That might indicate an issue

ceph_iscsi_config/device_status.py
```diff
@@ -0,0 +1,165 @@
...
+                        if dev_status.stable_cnt > settings.config.stable_state_reset_cnt:
+                            dev_status.stable_cnt = 0
+                            dev_status.state = "Online"
+                    else:
+                        dev_status.changed_state = False
+
+            # debugging info
+            dev_status = self.get_dev_status(image_name)
+            stats_dict = dev_status.get_status_dict()
+            self.logger.error(stats_dict)
```
dillaman, line 165: Nit: switch to debug?

ceph_iscsi_config/device_status.py
```diff
@@ -0,0 +1,165 @@
...
+            # there is valid data. We might not see it until the Nth call.
+            return
+
+        counter_dict = self.gw_counter_lookup.get(gw)
+        if counter_dict is None:
+            counter_dict = {}
+
+        for name, val in status.items():
+            if update_lock_owner is True and name == "lock_owner" and val == "true":
+                print("gw = {}".format(gw))
```
dillaman, line 70: Nit: use debug logging if required

ceph_iscsi_config/device_status.py
```diff
@@ -0,0 +1,165 @@
...
+                self.lock_owner = gw
+                continue
+
+            if name != "cmd_timed_out_cnt" and name != "conn_lost_cnt" and \
```
dillaman, line 74: Nit: 
```suggestion
            if name not in ["cmd_timed_out_cnt", "conn_lost_cnt", "lock_lost_cnt"]:
```

ceph_iscsi_config/device_status.py
```diff
@@ -0,0 +1,165 @@
...
+            if svc is None:
+                continue
+
+            tmax = None
+            image_names_dict = {}
+            for daemon, daemon_info in svc.items():
+                gw, image_name = daemon.split(":", 1)
+                image_names_dict[image_name] = image_name
+
+                t1 = datetime.strptime(daemon_info['status_stamp'], "%Y-%m-%dT%H:%M:%S.%f%z")
```
dillaman, line 133: Nit: I would just pass this update timestamp to `update_status` instead of the `update_lock_owner` param. Then if you store that timestamp (i.e. `self.lock_owner_timestamp`) the `update_status` method can perform that check directly to ensure it doesn't ping-pong due to stale statuses. Plus then you can also reset the `stable_cnt` if you detect that the owner was updated.

ceph_iscsi_config/device_status.py
lxbsz: Will fix it.

ceph_iscsi_config/device_status.py
lxbsz: Yeah, this makes sense.

lxbsz: BTW, when the `tcmu-runner` is not avaliable, the `status` output will be `{}`:

```
{
    "allocating_host": "lxbceph2",
    "backstore": "user:rbd",
    "backstore_object_name": "data.block01",
    "controls": {},
    "created": "2020/08/13 01:01:40",
    "image": "block01",
    "owner": "lxbceph1",
    "pool": "data",
    "pool_id": 9,
    "status": {},
    "updated": "2020/08/13 01:14:29",
    "wwn": "ad1a0c5c-a666-434b-90ab-dfc851ba5a02"
}
```

Will this be enough ?

gwcli/storage.py
lxbsz: At the same time I will rename the `current_owner` to `lock_owner`, which will be the same with the output from the curl GET. Or it will be a little confusing.

ceph_iscsi_config/device_status.py
lxbsz: Yeah, this make sense, and have fixed them all.

ceph_iscsi_config/gateway_setting.py
```diff
@@ -204,3 +204,8 @@ def normalize(self, raw_val):
     "qfull_timeout": IntSetting("qfull_timeout", 0, 600, 5),
     "osd_op_timeout": IntSetting("osd_op_timeout", 0, 600, 30),
     "hw_max_sectors": IntSetting("hw_max_sectors", 1, 8192, 1024)}
+
+TCMU_DEV_STATUS_SETTINGS = {
+    "lock_lost_cnt_thresh": IntSetting("lock_lost_cnt_thresh", 1, 1000000, 12),
+    "status_check_interval": IntSetting("status_check_interval", 1, 600, 60),
```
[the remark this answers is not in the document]
dillaman: Good point. Updates will only arrive from `tcmu-runner` to the `MGR` approximately every `mgr_stats_period` (5 seconds for the default). However, we also don't want to have lots of periodic commands being sent to the `MGR` asking for a status dump. I think 10 should be fine, however.

Also, right now because the watcher is using `mon_command` instead of `mgr_command`, the requests will be proxied via the `MON` to the `MGR`, so we might want to fix that to send directly to the `MGR`.

ceph_iscsi_config/device_status.py
```diff
@@ -0,0 +1,164 @@
...
+
+        for name, val in status.items():
+            if name == "lock_owner" and val == "true":
+                dt = datetime.strptime(status_stamp, "%Y-%m-%dT%H:%M:%S.%f%z")
+                if self.lock_owner_timestamp is None or dt > self.lock_owner_timestamp:
+                    self.lock_owner_timestamp = dt
+                    self.lock_owner = gw
+
+                    self.stable_cnt += 1
+                    if self.stable_cnt > settings.config.stable_state_reset_count:
```
dillaman, line 76: Nit: this is handled by the `run` method below

ceph_iscsi_config/device_status.py
```diff
@@ -0,0 +1,164 @@
...
+            counter_dict = {}
+
+        for name, val in status.items():
+            if name == "lock_owner" and val == "true":
+                dt = datetime.strptime(status_stamp, "%Y-%m-%dT%H:%M:%S.%f%z")
+                if self.lock_owner_timestamp is None or dt > self.lock_owner_timestamp:
+                    self.lock_owner_timestamp = dt
+                    self.lock_owner = gw
+
+                    self.stable_cnt += 1
```
dillaman, line 75: Nit: reset the stable_cnt to zero when the lock owner changes

ceph_iscsi_config/device_status.py
dillaman: Yeah, could output `Unknown` or similar on the CLI

ceph_iscsi_config/gateway_setting.py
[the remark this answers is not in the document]
lxbsz: Yeah, that makes sense, I will fix it.

ceph_iscsi_config/device_status.py
lxbsz: Sure, will fix it. Thanks.

lxbsz: Sure, will fix it.

ceph_iscsi_config/gateway_setting.py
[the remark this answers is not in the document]
lxbsz: Certainly that will make sense. Will fix it.

ceph_iscsi_config/device_status.py
```diff
@@ -0,0 +1,161 @@
...
+            for daemon, daemon_info in svc.items():
+                gw, image_name = daemon.split(":", 1)
+                image_names_dict[image_name] = image_name
+
+                dev_status = self.get_dev_status(image_name)
+                if dev_status is None:
+                    dev_status = TcmuDevStatusTracker(image_name)
+                    self.status_lookup[image_name] = dev_status
+
+                dev_status.update_status(gw, daemon_info.get('status'), daemon_info.get('status_stamp'))
```
dillaman, line 141: Tox: `./ceph_iscsi_config/device_status.py:141:100: E501 line too long (104 > 99 characters)`

ceph_iscsi_config/device_status.py
```diff
@@ -0,0 +1,161 @@
...
+        self.cluster.connect()
+
+        while True:
+            time.sleep(settings.config.status_check_interval)
+
+            cmd = json.dumps({"prefix": "service status", "format": "json"})
+
+            ret, outb, outs = self.cluster.mgr_command(cmd, b'')
+            if ret != 0:
+                self.logger.error("mon command failed {}".format(ret))
```
dillaman, line 122:
```suggestion
                self.logger.error("mgr command failed {}".format(ret))
```

rbd-target-api.py
```diff
@@ -1008,7 +1009,15 @@ def disk(pool, image):
...
 
         if image_id in config.config['disks']:
-            return jsonify(config.config["disks"][image_id]), 200
+            disk_dict = config.config["disks"][image_id]
+            global dev_status_watcher
+            disk_status = dev_status_watcher.get_dev_status(image_id)
+            if disk_status:
+                disk_dict['status'] = disk_status.get_status_dict()
+            else:
+                disk_dict['status'] = 'Unknown'
```
dillaman, line 1018: Nit: should this be assigning a dict to keep it consistent: 
```suggestion
                disk_dict['status'] = {'state': 'Unknown'}
```

ceph_iscsi_config/device_status.py
lxbsz: Sure, I will fix it. Didn't get this warning locally.

ceph_iscsi_config/device_status.py
lxbsz: Will fix it.

rbd-target-api.py
lxbsz: Sure.

sebastian-philipp:
FYI, this PR broke cephadm due to

https://github.com/ceph/ceph-iscsi/blob/b5355b8e61e58f3c4c1231f6a6f93dfc1bf1a156/ceph_iscsi_config/device_status.py#L118

See https://github.com/ceph/ceph/pull/38982

PR #216, opened 2020-09-08
Rename blacklist to blocklist

matthewoliver:
As a followup to Sage's ceph change[0], this patch changes:

  blacklist -> blocklist

The most important part is when we call `ceph osd blacklist {ls,rm}`
because it's now actaully `ceph osd blocklist ...`.

So this patch corrects this so ceph-iscsi will work in cephadm on
current master.

[0] - https://github.com/ceph/ceph/commit/dfd01d765304ed8783cef613930e65980d9aee23

### DIFF
--- a/ceph_iscsi_config/gateway.py
+++ b/ceph_iscsi_config/gateway.py
@@ -26,70 +26,88 @@ class CephiSCSIGateway(object):
         else:
             self.hostname = this_host()
 
-    def ceph_rm_blacklist(self, blacklisted_ip):
+    def _run_ceph_cmd(self, cmd, stderr=None, shell=True):
+        if not stderr:
+            stderr = subprocess.STDOUT
+        try:
+            result = subprocess.check_output(cmd, stderr=stderr, shell=shell)
+        except subprocess.CalledProcessError as err:
+            return None, err
+        return result, None
+
+    def ceph_rm_blocklist(self, blocklisted_ip):
         """
-        Issue a ceph osd blacklist rm command for a given IP on this host
-        :param blacklisted_ip: IP address (str - dotted quad)
+        Issue a ceph osd blocklist rm command for a given IP on this host
+        :param blocklisted_ip: IP address (str - dotted quad)
         :return: boolean for success of the rm operation
         """
 
-        self.logger.info("Removing blacklisted entry for this host : "
-                         "{}".format(blacklisted_ip))
+        self.logger.info("Removing blocklisted entry for this host : "
+                         "{}".format(blocklisted_ip))
 
         conf = settings.config
-        try:
-            subprocess.check_output("ceph -n {client_name} --conf {cephconf} "
-                                    "osd blacklist rm {blacklisted_ip}".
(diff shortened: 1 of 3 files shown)


matthewoliver:
Forgot to add the sage commit link to commit message.

lxbsz:
BTW, for this should we distinguish whether we are using the the old ceph version for the `osd blacklist rm/ls...` ? Or it won't work with the old ceph version.

And the we have the same issue with the tcmu-runner's rbd handler, so I am still waiting to know how to make it compitble with both.

lxbsz:
> BTW, for this should we distinguish whether we are using the the old ceph version for the `osd blacklist rm/ls...` ? Or it won't work with the old ceph version.
> 
> And the we have the same issue with the tcmu-runner's rbd handler, so I am still waiting to know how to make it compitble with both.

Okay, my bad, sage's patch have already done this. Please ignore this.

ceph_iscsi_config/gateway.py
```diff
@@ -26,46 +26,46 @@ def __init__(self, logger, config, name=None):
...
-                                    "osd blacklist rm {blacklisted_ip}".
-                                    format(blacklisted_ip=blacklisted_ip,
+                                    "osd blocklist rm {blocklisted_ip}".
+                                    format(blocklisted_ip=blocklisted_ip,
                                            client_name=conf.cluster_client_name,
                                            cephconf=conf.cephconf),
                                     stderr=subprocess.STDOUT, shell=True)
         except subprocess.CalledProcessError as err:
-            self.logger.critical("blacklist removal failed: {}. Run"
+            self.logger.critical("blocklist removal failed: {}. Run"
```
dillaman, line 48: You will need to retry using the "osd blacklist rm" command if this throws to ensure backwards compatibility

ceph_iscsi_config/gateway.py
```diff
@@ -74,22 +74,22 @@ def osd_blacklist_cleanup(self):
...
-            blacklist = subprocess.check_output("ceph -n {client_name} --conf {cephconf} "
-                                                "osd blacklist ls"
+            blocklist = subprocess.check_output("ceph -n {client_name} --conf {cephconf} "
+                                                "osd blocklist ls"
                                                 .format(client_name=conf.cluster_client_name,
                                                         cephconf=conf.cephconf),
                                                 shell=True,
                                                 stderr=subprocess.STDOUT)
         except subprocess.CalledProcessError:
             self.logger.critical("Failed to run 'ceph -n {client_name} --conf {cephconf} "
```
dillaman, line 84: Nit: similar issue here

dillaman (changes requested):

matthewoliver:
Maybe something like this then.

Still want to test it.

ceph_iscsi_config/gateway.py
matthewoliver: done

dillaman (approved):
lgtm

smithfarm:
This fixes the cephadm iscsi deployment breakage in master, at least for me! Yay.

PR #217, opened 2020-10-02
Make settings mon config key store aware

matthewoliver:
When we are dealing with things like `trusted_ip_list` but ceph-iscsi is
running in a container modifying the list can be problematic. Sure the
config file is mounted into the container, but getting users to modify
that isn't very user friendly.

What would be better is if you could store the value in the mon config
key store. But then why stop at just the `trusted_ip_list`, ceph-iscsi's
settings are already fairly dynamic. So instead treat any value starting
with `config://` as a key for config key store.

This obviously wont work for the client name or the ceph conffile as we
need those to talk to the key store, but means anything else in the
config, if you so choose, can come from the dynamic ceph config store.

For example, if I had:

  trusted_ip_list = config://some/key/you/want/to/use

Then it'll look in the ceph config store under the key
`some/key/you/want/to/use`.

To set the key you just need to:

  ceph config-key set some/key/you/want/to/use "192.168.0.1,192.168.0.2"

And then restart the container.

### DIFF
--- a/ceph_iscsi_config/settings.py
+++ b/ceph_iscsi_config/settings.py
@@ -8,6 +8,7 @@ except ImportError:
 
 import hashlib
 import json
+import rados
 import re
 
 from ceph_iscsi_config.gateway_setting import (TGT_SETTINGS, SYS_SETTINGS,
@@ -21,6 +22,9 @@ from ceph_iscsi_config.gateway_setting import (TGT_SETTINGS, SYS_SETTINGS,
 def init():
     global config
     config = Settings()
+
+
+MON_CONFIG_PREFIX = 'config://'
 
 
 class Settings(object):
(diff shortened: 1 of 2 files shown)


matthewoliver:
I've tested it by moving a few different params to the config store. The trusted_ip_list (a list), api port (an int) and api user (str). They all in essense come in initally as strings anyway, or rather bytestrings. But so far, they seems to all be getting their config-key value equivs.

ceph_iscsi_config/settings.py
```diff
@@ -23,6 +24,9 @@ def init():
     config = Settings()
 
 
+MON_CONFIG_PREFIX = 'config://'
```
dillaman, line 27: Nit: perhaps just to be more explicit about the source:
```suggestion
MON_CONFIG_PREFIX = 'mon-config-key://'
```

matthewoliver: Yeah, I agree that 'config://' isn't the most explicit or obvious prefix. Only chose it as it was already chosen in RGW to do the exact same thing, so thought keeping things same across the project would make it clearer and easier to operate.

Happy to have my arm twisted though ;)

dillaman: Ah -- didn't realize it was in RGW. If that's the case, objection withdrawn

dillaman (approved):
:+1:

Issue #219, opened 2020-10-14
 rbd-target-api's setting `trusted_ip_list` can't contain spaces


bk201:
The expected trusted IPs are not correct if `trusted_ip_list` setting contains spaces.

e.g. space after comma sign:

```
trusted_ip_list = 192.168.0.1, 192.168.0.2, 192.168.0.3

# -> parsed as ['192.168.0.1', ' 192.168.0.2', ' 192.168.0.3'], the later two contain prefixed spaces.
``` 

e.g. no spaces.

```
trusted_ip_list = 192.168.0.1,192.168.0.2,192.168.0.3

# -> parsed as ['192.168.0.1', '192.168.0.2', '192.168.0.3'], which is correct.
``` 

On possible solution is to strip the space when normalizing: 
https://github.com/ceph/ceph-iscsi/blob/46ba8e554e0d87ce516c10d576956f09660997b3/ceph_iscsi_config/gateway_setting.py#L73-L74

PR #220, opened 2020-10-14
Add a strip to ListSetting.Normalize

matthewoliver:
Currently when you specify a list in ceph-iscsi it _MUST_ be in the form
of 'item1,item2,item3'. Just delimitted by a comma (,). So if you
specify trusted_ip_list with spaces:

  trusted_ip_list: '10.1.1.1, 10.1.1.2'

This is actually fail as the space before 10.1.1.2 is retained.

This patch adds a strip() to the normalise call removing any extra
whitespace from the start and end of each item in the list.

Fixes: https://github.com/ceph/ceph-iscsi/issues/219


dillaman (approved):
lgtm

PR #228, opened 2020-12-04
iscsi: raise if the 'gateway.conf' config file doesn't exist

lxbsz:
The Config class will create a empty 'gateway.conf' object at the
begining if it doesn't exist. It should be buggy if it doesn't
exist here.


dillaman (approved):
lgtm

PR #229, opened 2020-12-10
rbd-target-api: misc fixing for disk API

lxbsz:
Fixes: https://tracker.ceph.com/issues/48541

### DIFF
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -1129,7 +1129,7 @@ def disk(pool, image):
 
         api_vars = {
             'purge_host': local_gw,
-            'preserve_image': request.form['preserve_image'],
+            'preserve_image': request.form.get('preserve_image'),
             'backstore': backstore
         }
 
@@ -1272,7 +1272,7 @@ def _disk(pool, image):
 
         # if valid_request(request.remote_addr):
         purge_host = request.form['purge_host']
-        preserve_image = request.form['preserve_image'] == 'true'
+        preserve_image = request.form.get('preserve_image') == 'true'
         logger.debug("delete request for disk image '{}'".format(image_id))
         pool, image = image_id.split('/', 1)
         disk_config = config.config['disks'][image_id]
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -989,15 +989,15 @@ def disk(pool, image):
     :param count: (str) the number of images will be created
     :param owner: (str) the owner of the rbd image
     :param controls: (JSON dict) valid control overrides
-    :param preserve_image: (bool) do NOT delete RBD image
-    :param create_image: (bool) create RBD image if not exists
+    :param preserve_image: (bool, 'true/false') do NOT delete RBD image
+    :param create_image: (bool, 'true/false') create RBD image if not exists, true as default
     :param backstore: (str) lio backstore
     :param wwn: (str) unit serial number
     **RESTRICTED**
     Examples:
     curl --user admin:admin -d mode=create -d size=1g -d pool=rbd -d count=5
         -X PUT http://192.168.122.69:5000/api/disk/rbd/new0_
-    curl --user admin:admin -d mode=create -d size=10g -d pool=rbd -d create_image=true
+    curl --user admin:admin -d mode=create -d size=10g -d pool=rbd -d create_image=false
         -X PUT http://192.168.122.69:5000/api/disk/rbd/new1
     curl --user admin:admin -X GET http://192.168.122.69:5000/api/disk/rbd/new2
     curl --user admin:admin -X DELETE http://192.168.122.69:5000/api/disk/rbd/new3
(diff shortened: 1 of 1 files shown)


rbd-target-api.py
```diff
@@ -992,11 +992,11 @@ def disk(pool, image):
     **RESTRICTED**
     Examples:
     curl --insecure --user admin:admin -d mode=create -d size=1g -d pool=rbd -d count=5
-        -X PUT https://192.168.122.69:5000/api/disk/rbd.new2_
-    curl --insecure --user admin:admin -d mode=create -d size=10g -d pool=rbd
-        -X PUT https://192.168.122.69:5000/api/disk/rbd.new3_
-    curl --insecure --user admin:admin -X GET https://192.168.122.69:5000/api/disk/rbd.new2_1
-    curl --insecure --user admin:admin -X DELETE https://192.168.122.69:5000/api/disk/rbd.new2_1
+        -X PUT https://192.168.122.69:5000/api/disk/rbd/new0
```
dillaman, line 995: Nit: I think the rationale here was that `count=5` so it's going to create 5 disks and append the index at the end of the image name

rbd-target-api.py
```diff
@@ -992,11 +992,11 @@ def disk(pool, image):
...
+    curl --insecure --user admin:admin -d mode=create -d size=10g -d pool=rbd -d create_image=true
```
dillaman, line 996: Nit: `create_image` is implied if `size` is provided -- a better example would be to drop `size` and set `create_image=false`

lxbsz: It seems not to the current code:

```
# curl --insecure --user admin:admin -d mode=create -d size=100m -d pool=rbd2 -d image=block1 -d count=1 -X PUT http://192.168.195.172:5000/api/disk/rbd2/block1
{"message":"Image rbd2/block1 does not exist"}

```

rbd-target-api.py
lxbsz: Okay, will add it back.

rbd-target-api.py
lxbsz: The code:

`if mode == 'create' and (not create_image or not size):`

Nit: the `if` branch won't create the image. That means only when `size` is specified and the `create_image` must be `true` will it create the image.

dillaman: Thx -- but does that mean the previous example is also wrong since it doesn't have `create_image=true` (given the image name of `new...` implies it should be created not reused).

lxbsz: The logic changed a lot, I have test the REST APIs one by one 3 years ago and the example was correct then. I will try to check and fix them if there has issues.

dillaman: Ack -- so it might be easier to leave this PR open as a running collection of fixes as you go through developing the as-built tests.

lxbsz: Sure.

lxbsz:
@dillaman This is ready to review now. Thanks.

rbd-target-api.py
```diff
@@ -229,7 +229,7 @@ def get_targets():
     List targets defined in the configuration.
     **RESTRICTED**
     Examples:
-    curl --insecure --user admin:admin -X GET https://192.168.122.69:5000/api/targets
+    curl --insecure --user admin:admin -X GET http://192.168.122.69:5000/api/targets
```
dillaman, line 232: Nit: by changing "https" -> "http", that `--insecure` optional is no longer required (throughout) so perhaps just drop it.

lxbsz: I am afraid it won't correct, the default it using the https:

`"api_secure": BoolSetting("api_secure", True),`
`http_mode = 'https' if settings.config.api_secure else 'http'`

lxbsz: > I am afraid it is not correct, the default is using the https:
> 
> `"api_secure": BoolSetting("api_secure", True),`
> `http_mode = 'https' if settings.config.api_secure else 'http'`

Sorry, please ignore this. It works.

lxbsz: Fixed it.

rbd-target-api.py
```diff
@@ -1069,17 +1069,26 @@ def disk(pool, image):
...
             try:
+                # not size meaning not intend to creat an image, try to
+                # check whether it exists
                 rbd_image = RBDDev(image, 0, backstore, pool)
                 size = rbd_image.current_size
             except rbd.ImageNotFound:
-                if not create_image:
-                    return jsonify(message="Image {} does not exist".format(image_id)), 400
-                else:
+                # the create_image=true will be implied if size is spcified
```
dillaman, line 1084: `spcified` -> `specified`

rbd-target-api.py
```diff
@@ -1069,17 +1069,26 @@ def disk(pool, image):
...
-        create_image = request.form.get('create_image') == 'true'
-        if mode == 'create' and (not create_image or not size):
+        create_image = request.form.get('create_image', 'true')
+        if create_image not in ['true', 'false']:
+            logger.error("Invalid 'create_image' value {}".format(create_image))
+            return jsonify(message="Invalid 'create_image' value {}".format(create_image)), 400
+
+        if mode == 'create' and (create_image == 'false' or not size):
             try:
+                # not size meaning not intend to creat an image, try to
```
dillaman, line 1079: `not size meaning not intend to creat an image, try to` -> `no size implies not intention to create an image, try to

rbd-target-api.py
lxbsz: Fixed.

rbd-target-api.py
lxbsz: Done.

dillaman (approved):
lgtm

lxbsz:
@dillaman BTW, how to trigger to build the ceph-iscsi package ?

dillaman:
The builds are triggered automatically whenever you push a `wip-XYZ` branch to this repo or whenever a new PR merges to `master` Here is the latest [1][2].

[1] https://jenkins.ceph.com/job/ceph-iscsi/478/
[2] https://shaman.ceph.com/repos/ceph-iscsi/master/b5355b8e61e58f3c4c1231f6a6f93dfc1bf1a156/

lxbsz:
> The builds are triggered automatically whenever you push a `wip-XYZ` branch to this repo or whenever a new PR merges to `master` Here is the latest [1][2].
> 
> [1] https://jenkins.ceph.com/job/ceph-iscsi/478/
> [2] https://shaman.ceph.com/repos/ceph-iscsi/master/b5355b8e61e58f3c4c1231f6a6f93dfc1bf1a156/

Okay, get it. Thanks.

PR #235, opened 2021-06-03
README: add IPv6 related info

lxbsz:

### DIFF
--- a/README
+++ b/README
@@ -12,9 +12,9 @@ configure multiple gateways concurrently.
 
 Here's an example of the shell interface the gwcli tool provides:
 
-/> ls
+[ceph-iscsi]$ gwcli ls
 o- / .................................................................................. [...]
-  o- clusters ................................................................. [Clusters: 1]
+  o- cluster .................................................................. [Clusters: 1]
   | o- ceph ..................................................................... [HEALTH_OK]
   |   o- pools ................................................................... [Pools: 3]
   |   | o- ec ........................................ [(2+1), Commit: 0b/40G (0%), Used: 0b]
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -60,13 +70,31 @@ add the required variables to the body section in the client ui
...
      -X PUT https://192.168.122.69:5000/api/gateway/rh7-gw2
 
+IPv6 Supporting:
+Ceph-iscsi has supported the IPv6 already, for this to make sure the IPv6
+addresses used are global unicast ones, and then the URL link could be:
+
+curl --insecure --user admin:admin -d "ip_address=2006:ac81::1104" \
+     -X PUT https://[2006:ac81::1104]:5000/api/gateway/rh7-gw2
+
+And mix using the IPv4 and IPv6 is also allowed as bellow:
```
idryomov, line 87:
```suggestion
And mixing the IPv4 and IPv6 addresses is also allowed as below:
```

README
```diff
@@ -60,13 +70,31 @@ add the required variables to the body section in the client ui
...
+
+curl --insecure --user admin:admin -d "ip_address=192.168.122.104" \
+     -X PUT https://[2006:ac81::1104]:5000/api/gateway/rh7-gw2
+
+curl --insecure --user admin:admin -d "ip_address=2006:ac81::1104" \
+     -X PUT https://192.168.122.104:5000/api/gateway/rh7-gw2
```
idryomov, line 93: Since you changed the example `gwcli` output, can you make the IPs consistent?  In particular, `192.168.122.104` is no longer there.

lxbsz: Fixed them all. Thanks.

README
```diff
@@ -60,13 +70,31 @@ add the required variables to the body section in the client ui
...
-curl --insecure --user admin:admin -d "ip_address=192.168.122.104" \
+curl --insecure --user admin:admin -d "ip_address=192.168.122.14" \
      -X PUT https://192.168.122.69:5000/api/gateway/rh7-gw2
 
+IPv6 Support:
+Make sure the IPv6 addresses used are global unicast ones, and then the
+URL link could be:
+
+curl --insecure --user admin:admin -d "ip_address=2006:ac81::1104" \
+     -X PUT https://[2006:ac81::1104]:5000/api/gateway/rh7-gw2
```
idryomov, line 85: In this example we are asking `rh7-gw1` to define a second gateway named `rh7-gw2` at that IP address, correct?  If so, should it be
```suggestion
     -X PUT https://[2006:ac81::1103]:5000/api/gateway/rh7-gw2
```
and similarly below?

idryomov: And it looks like the proper route is `/api/gateway/<target_iqn>/<gateway_name>` with both `target_iqn` and `gateway_name` required.

Can you test the updated examples to make sure they actually work?

lxbsz: Without the `<target_iqn>` is the old style. I will copy this from my cmd history.

lxbsz: > In this example we are asking `rh7-gw1` to define a second gateway named `rh7-gw2` at that IP address, correct? If so, should it be
> 

This is correct !

The IPv6 address `2006:ac81::1104` is in rh7-gw2 node. This command means it is requesting the `rh7-gw2` node to set up the gateway for `rh7-gw2` node.

> and similarly below?

idryomov: I'm confused.  If there is no gateway on `rh7-gw2` node yet, how can one request anything from it?

lxbsz: > I'm confused. If there is no gateway on `rh7-gw2` node yet, how can one request anything from it?
> 

Checked the history commits about this change, for the old style the `<target_iqn>` was one argument instead. It seems someone made one mistake when changing the style, just deleted the `<target_iqn>` argument in the examples but forgot to add it in the URL path. 


> I'd expect `curl` connect to `rh7-gw1` (`::1103`) and provide `rh7-gw2` IP address (`::1104`) in `ip_address` argument. Sorry for being dense on a simple example but I want to understand what I am missing.

Okay, there is no limitation about which one node should be connected when creating gateways except the first gateway to be created must be the same node it is connecting to.

idryomov (approved):

PR #239, opened 2021-07-04
gwcli: fix api_endpoint AttributeError

lxbsz:
AttributeError: 'Settings' object has no attribute 'api_endpoint'

### DIFF
--- a/gwcli.py
+++ b/gwcli.py
@@ -105,7 +105,7 @@ def main():
     root_node.refresh()
     if root_node.error:
         print("Unable to contact the local API endpoint "
-              "({})".format(settings.config.api_endpoint))
+              "({})".format(root_node.local_api))
         sys.exit(-1)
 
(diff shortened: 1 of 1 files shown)


idryomov (approved):

gwcli.py
```diff
@@ -105,7 +105,7 @@ def main():
     root_node.refresh()
     if root_node.error:
         print("Unable to contact the local API endpoint "
-              "({})".format(settings.config.api_endpoint))
+              "({})".format(settings.config.local_api))
```
idryomov, line 108: Wait, shouldn't it be `self.local_api`?

lxbsz: It's root_node.local_api. Fixed.

lxbsz: [root@node02 ~]# gwcli  ls
Unable to contact the local API endpoint (http://localhost:5000/api)

PR #241, opened 2021-07-23
change blocklist to blacklist.

locallocal:
Hi,I don‘t know it is misspelling or deliberate.but I think it's **blacklist**.


lxbsz:
> Hi,I don‘t know it is misspelling or deliberate.but I think it's **blacklist**.

This is deliberate and for the old ceph version it was using the `blacklist` and now all are switched to `blocklist`.

More detail please see https://github.com/ceph/ceph/pull/36527

locallocal:
@lxbsz Thx.

PR #250, opened 2021-12-14
spec: mark the ceph-iscsi-tools package deprecated

lxbsz:
Ceph cluster now incorporates a generic metric gathering framework
within the OSDs and MGRs to provide built-in monitoring.

### DIFF
--- a/ceph-iscsi.spec
+++ b/ceph-iscsi.spec
@@ -36,6 +36,7 @@ BuildArch:      noarch
 
 Obsoletes:      ceph-iscsi-config
 Obsoletes:      ceph-iscsi-cli
+Obsoletes:      ceph-iscsi-tools
 
 Requires:       tcmu-runner >= 1.4.0
 Requires:       ceph-common >= 10.2.2


idryomov:
I'd update the commit message to say that otherwise an upgrade of ceph-iscsi on a system with ceph-iscsi-tools installed would fail (just to make it clear to folks who are not familiar with RPM and spec files).

idryomov (approved):

lxbsz:
> I'd update the commit message to say that otherwise an upgrade of ceph-iscsi on a system with ceph-iscsi-tools installed would fail (just to make it clear to folks who are not familiar with RPM and spec files).

Sure, thanks. I will backport this later this week.

lxbsz:
> I'd update the commit message to say that otherwise an upgrade of ceph-iscsi on a system with ceph-iscsi-tools installed would fail (just to make it clear to folks who are not familiar with RPM and spec files).

Sorry I misreaded this and triggered the merge button :-(.

idryomov:
No problem -- I should have just left a comment instead of approving right away.

PR #253, opened 2022-02-23
Add initial unit tests for gateway

gnuoy:
Add some intial unit tests to gateway.py


test/test_gateway.py
```diff
@@ -0,0 +1,110 @@
...
+        self.gw = gateway.CephiSCSIGateway(
+            self.logger,
+            self.settings,
+            'thishost')
+
+    def mock_run_ceph_cmd(self, cmd, stderr=None, shell=True):
+        return self.run_ceph_mock_return.pop()
+
+    @mock.patch.object(gateway.subprocess, 'check_output')
+    def test__run_ceph_cmd(self, check_output):
```
lxbsz, line 28: s/test__run_ceph_cmd/test_run_ceph_cmd/ ?

gnuoy: The method being tested starts with an underscore `_run_ceph_cmd` . Hence the double underscore in the test name. I'm happy to change the test name though if you'd prefer `test_run_ceph_cmd`

lxbsz: Yeah, please.

lxbsz: At the same time could update the `README` to give some explaination and a example for the test ?

gnuoy: Hi @lxbsz, thank you for the review. I can't think what would be relevant to put in the README since this PR is just adding some unit tests or are you suggesting I add doc strings to the tests?

lxbsz: I am how to run the test. You need to update the exact command and what should others do to run the test.

Such as the `Curl Examples:`.

test/test_gateway.py
```diff
@@ -25,7 +25,7 @@ def mock_run_ceph_cmd(self, cmd, stderr=None, shell=True):
         return self.run_ceph_mock_return.pop()
 
     @mock.patch.object(gateway.subprocess, 'check_output')
-    def test__run_ceph_cmd(self, check_output):
+    def test_run_ceph_cmd(self, check_output):
```
lxbsz, line 28: Could you squash this commit into the first one ?

test/test_gateway.py
gnuoy: I thought the README was aimed at users who want to install, configure and use ceph-iscsi. The `Curl Examples` in the README is aimed at users who want to interact with the api via curl. Running unit tests seems to me to only be of interest to developers. Having said that I could add something like 

```
diff --git a/README b/README
index fe32bd2..b0c7455 100644
--- a/README
+++ b/README
@@ -135,3 +135,9 @@ here:
 http://docs.ceph.com/docs/master/rbd/iscsi-target-cli/
 
 can be used to create a iscsi-gateway.cfg and create a target.
+
+## Running Unit Tests
+
+The unit tests for ceph-isci can be run with:
+
+```tox -e py3```
```

Is that the sort of thing you are looking for?

PR #256, opened 2022-03-17
Improve werkzeug version checking

### LINKED ISSUE #255: rbd-target-api will not start on Ubuntu Impish and later
The rbd-target-api fails to start on Ubuntu Impish (21.10) and later. This appears to be caused by a werkzeug package revision check in rbd-target-api. The check is used to decide whather to add an OpenSSL.SSL.Context or a ssl.SSLContext. The code comment suggests that ssl.SSLContext is used for werkzeug 0.9 so that TLSv1.2 can be used. It is also worth noting that support for OpenSSL.SSL.Context was dropped in werkzeug 0.10. The intention of the check appears to be to add OpenSSL.SSL.Context if the version of is werkzeug is below 0.9  otherwise use ssl.SSLContext. When rbd-target-api checks the werkzeug revision it only looks at the minor revision number and Ubuntu Impish contains werkzeug 1.0.1 which obviously has a minor revision number of 0 which causes rbd-target-api to use an OpenSSL.SSL.Context which is not supported by werkzeug which causes:

```
# /usr/bin/rbd-target-api
 * Serving Flask app 'rbd-target-api' (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
  File "/usr/bin/rbd-target-api", line 3022, in <module>
    main()
  File "/usr/bin/rbd-target-api", line 2952, in main
    app.run(host=settings.config.api_host,
  File "/usr/lib/python3/dist-packages/flask/app.py", line 922, in run
    run_simple(t.cast(str, host), port, self, **options)
  File "/usr/lib/python3/dist-packages/werkzeug/serving.py", line 1010, in run_simple
    inner()
  File "/usr/lib/python3/dist-packages/werkzeug/serving.py", line 950, in inner
    srv = make_server(
  File "/usr/lib/python3/dist-packages/werkzeug/serving.py", line 782, in make_server
    return ThreadedWSGIServer(
  File "/usr/lib/python3/dist-packages/werkzeug/serving.py", line 708, in __init__
    self.socket = ssl_context.wrap_socket(self.socket, server_side=True)
AttributeError: 'Context' object has no attribute 'wrap_socket'
```

gnuoy:
Check the werkzeug major version as well as the minor version when
deciding which type of SSL context. Closes #255

### DIFF
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -2879,7 +2879,7 @@ def get_ssl_context():
             return None
 
     ver, rel, mod = werkzeug.__version__.split('.')
-    if int(rel) > 9:
+    if int(ver) > 0 or int(rel) > 9:
         logger.info("API server using TLSv1.2")
 
         context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)


trociny (approved):
LGTM

idryomov (approved):

PR #260, opened 2022-05-30
README: add doc for mutual chap authentication

lxbsz:

### DIFF
--- a/README
+++ b/README
@@ -12,7 +12,7 @@ configure multiple gateways concurrently.
 
 Here's an example of the shell interface the gwcli tool provides:
 
-[ceph-iscsi]$ gwcli ls
+$ gwcli ls
 o- / .................................................................................. [...]
   o- cluster .................................................................. [Clusters: 1]
   | o- ceph ..................................................................... [HEALTH_OK]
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -74,27 +74,49 @@ Curl Examples:
...
 
+Mutual CHAP Authentication:
+
+In the initiator node, uncomment and set the following options in /etc/iscsi/iscsid.conf:
+
+node.session.auth.authmethod = CHAP
+node.session.auth.username = ${admin_username}
+node.session.auth.password = ${admin_password}
+node.session.auth.username_in = ${mutual_username}
+node.session.auth.password_in = ${mutual_password}
```
idryomov, line 106: I think this should be added to the Ceph docs where the CHAP authentication is already touched on, so that it stays in one place.  Do you mind if I do that based on this PR?

README
```diff
@@ -74,27 +74,49 @@ Curl Examples:
...
+
+$ gwcli iscsi-targets/iqn.2003-01.com.redhat.iscsi-gw:ceph-igw0/hosts/iqn.1994-05.com.redhat:client/ auth \
+      username=${admin_username} password=${admin_password} mutual_username=${mutual_username} \
+      mutual_password=${mutual_password}
+
+Note that after you set some setting in /etc/iscsi/iscsid.conf you have to re-run the discovery command
+so those settings are used for the targets/portals returned from the target.
+
+And also please be aware that do not use the same values for both admin username/password and mutual
+username/password, and the validate characters for the username/password are [0~9,A~Z,a~z,_-@/|].
```
idryomov, line 119: Where is the set of valid characters defined?  Do you have a reference?

README
lxbsz: Sounds good, then let's paste that link in the `README` here ? At least we know where to find that doc and how to set it.

idryomov: https://github.com/ceph/ceph/pull/46426

README
lxbsz: More detail please see https://github.com/ceph/ceph-iscsi/blob/main/gwcli/utils.py#L172.

README
idryomov: > Sounds good, then let's paste that link in the `README` here ? At least we know where to find that doc and how to set it.

It is already there, at the bottom.

lxbsz:
This will be updated in the ceph doc, and closing this PR.

PR #263, opened 2022-08-26
disable JSONIFY_PRETTYPRINT_REGULAR in flask app config

lxbsz:
The request.is_xhr method has been deprecated & removed, as it was
unreliable.

### DIFF
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -43,6 +43,8 @@ from gwcli.utils import (APIRequest, valid_gateway, valid_client,
                          GatewayAPIError)
 
 app = Flask(__name__)
+# workaround for https://github.com/pallets/flask/issues/2549
+app.config['JSONIFY_PRETTYPRINT_REGULAR'] = False
 
 
 def requires_basic_auth(f):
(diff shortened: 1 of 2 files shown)


rbd-target-api.py
```diff
@@ -43,6 +43,7 @@
                          GatewayAPIError)
 
 app = Flask(__name__)
+app.config['JSONIFY_PRETTYPRINT_REGULAR'] = False
```
idryomov, line 46:
```suggestion

# workaround for https://github.com/pallets/flask/issues/2549
app.config['JSONIFY_PRETTYPRINT_REGULAR'] = False
```
and same for the Prometheus instance on the other file.

idryomov (commented):
I don't understand why this issue is being hit now.  Since we are using distro packages, does it mean that `python3-werkzeug` package got upgraded to >= 1.0.0. but `python3-flask` somehow did not?

lxbsz:
> I don't understand why this issue is being hit now. Since we are using distro packages, does it mean that `python3-werkzeug` package got upgraded to >= 1.0.0 but `python3-flask` somehow did not? Flask took care of it in 2017 in [pallets/flask#2193](https://github.com/pallets/flask/pull/2193)...

Not very sure, it seems due to some reason the distro was using the versions that the `is_xhr` is just marked as deprecated but still there and for recent release switched to `werkzeug >= 1.0.0`.

PR #264, opened 2022-08-26
lun: fix valid_disk() return value

idryomov:
valid_disk() should return either "ok" or an error description
string.  However, in case controls normalization fails, ValueError
object is returned.

Signed-off-by: Ilya Dryomov <idryomov@gmail.com>


lxbsz (approved):
LGTM.

PR #271, opened 2023-06-29
spec: make sure we always have the target dir

lxbsz:
rtslib now requires /etc/target or /var/target to be setup (it depends on the version). The issue is that rtslib does not make those dirs and instead relies on targetcli or ceph-iscs to make them.

For container cases we do not install targetcli, so ceph-iscsi's daemons will crash or fail to setup disks, when we make rtslib calls that try to access the etc/var target dirs.

When rtslib is fixed drop this and update the dep. Note that we add both etc and var, because the kernel and userspace developers keep switching the dir they want to use.

### DIFF
--- a/ceph-iscsi.spec
+++ b/ceph-iscsi.spec
@@ -21,7 +21,7 @@
 
 Name:           ceph-iscsi
 Version:        3.6
-Release:        1%{?dist}
+Release:        2%{?dist}
 Group:          System/Filesystems
 Summary:        Python modules for Ceph iSCSI gateway configuration management
 %if 0%{?suse_version}
@@ -133,6 +133,13 @@ mkdir -p %{buildroot}%{_mandir}/man8
 install -m 0644 gwcli.8 %{buildroot}%{_mandir}/man8/
 gzip %{buildroot}%{_mandir}/man8/gwcli.8
 mkdir -p %{buildroot}%{_unitdir}/rbd-target-gw.service.d
+# When rtslib is fixed drop this and update the dep. Note that we add both
+# /etc and /var, because the kernel and userspace developers keep switching
+# the dir they want to use.
+mkdir -p %{buildroot}%{_sysconfdir}/target/pr
+mkdir -p %{buildroot}%{_sysconfdir}/target/alua
+mkdir -p %{buildroot}%{_localstatedir}/target/pr
+mkdir -p %{buildroot}%{_localstatedir}/target/alua
 %if 0%{?suse_version}
 mkdir -p %{buildroot}%{_sbindir}
 ln -s service %{buildroot}%{_sbindir}/rcrbd-target-gw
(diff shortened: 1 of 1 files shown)


ceph-iscsi.spec
```diff
@@ -133,6 +133,13 @@ mkdir -p %{buildroot}%{_mandir}/man8
 install -m 0644 gwcli.8 %{buildroot}%{_mandir}/man8/
 gzip %{buildroot}%{_mandir}/man8/gwcli.8
 mkdir -p %{buildroot}%{_unitdir}/rbd-target-gw.service.d
+# When rtslib is fixed drop  this and update the dep. Note that we add both
+# etc and var, because the kernel and userspace developers keep switching
```
idryomov, line 137:
```suggestion
# When rtslib is fixed drop this and update the dep. Note that we add both
# /etc and /var, because the kernel and userspace developers keep switching
```

idryomov:
> The issue is that rtslib does not make those dirs and instead relies on targetcli or ceph-iscs to make them.

Typo: ceph-iscsi

> For container cases we do not install targetcli, so ceph-iscsi's daemons will crash or fail to setup disks, when we make rtslib calls that try to access the etc/var target dirs.

ceph-container doesn't mount `/etc/target` and/or `/var/target` from the host either, so I suspect the only reason it happens to work when there directories are created during package installation is that ceph-iscsi doesn't need to communicate with the kernel.  This get us past the checks in rtslib but no meaningful ceph-iscsi <-> kernel communication can occur because they would be reading from/writing to different locations (container vs host).

ceph-iscsi.spec
lxbsz: Sure, I just copied it from downstream. will fix it.

ceph-iscsi.spec
```diff
@@ -133,6 +133,13 @@ mkdir -p %{buildroot}%{_mandir}/man8
 install -m 0644 gwcli.8 %{buildroot}%{_mandir}/man8/
 gzip %{buildroot}%{_mandir}/man8/gwcli.8
 mkdir -p %{buildroot}%{_unitdir}/rbd-target-gw.service.d
+# When rtslib is fixed drop  this and update the dep. Note that we add both
```
idryomov, line 136:
```suggestion
# When rtslib is fixed drop this and update the dep. Note that we add both
```

lxbsz: Done.

idryomov:
Additional note: since https://github.com/open-iscsi/rtslib-fb/commit/99072f0cee35e1591f76b1c470a57f5299e2e527 rtslib is more aggressive about changing dbroot to what it likes better (upstream it's the "preferred" `/etc/target` which aligns with LIO in the kernel; however in Fedora, CentOS and RHEL it's the "default" `/var/target`).  This can cause trouble on live upgrades:

1) rtslib attempts to change dbroot upon restart of ceph-iscsi
2) kernel rejects the change because dbroot is in use (i.e. there are registered devices from before the upgrade)
3) rtslib checks whether the directory it attempted to change dbroot to exists, sees that it doesn't exist, assumes _that_ to be reason for the rejection and throws an exception

https://github.com/open-iscsi/rtslib-fb/blob/745d51a46e6718e34c59b728fd96ec08bbf906dc/rtslib/root.py#L169-L182

idryomov (approved):

PR #276, opened 2024-04-18
rbd-target-api: switch to use distro to get the release version

lxbsz:
The 'platform.linux_distribution()' has been deprecated since python 3.7. And we should switch to use the distro.version(), distro.id() and distro.name() instead.

### DIFF
--- a/ceph-iscsi.spec
+++ b/ceph-iscsi.spec
@@ -43,6 +43,7 @@ Requires:       ceph-common >= 10.2.2
 %if 0%{?with_python2}
 BuildRequires:  python2-devel
 BuildRequires:  python2-setuptools
+Requires:       python2-distro
 Requires:       python-rados >= 10.2.2
 Requires:       python-rbd >= 10.2.2
 Requires:       python-netifaces >= 0.10.4
@@ -60,6 +61,7 @@ Requires:       python2-requests
 %else
 BuildRequires:  python3-devel
 BuildRequires:  python3-setuptools
+Requires:       python3-distro
 Requires:       python3-rados >= 10.2.2
 Requires:       python3-rbd >= 10.2.2
 Requires:       python3-netifaces >= 0.10.4
--- a/gwcli/client.py
+++ b/gwcli/client.py
@@ -259,7 +259,7 @@ class Clients(UIGroup):
                     status = False
                     break
 
-        return "Auth: {}, Hosts: {}".format(auth_stat_str, len(self.children)),\
+        return "Auth: {}, Hosts: {}".format(auth_stat_str, len(self.children)), \
             status
 
 
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -2184,7 +2184,7 @@ def client(target_iqn, client_iqn):
                                         http_method='put',
                                         api_vars=api_vars)
 
-        return jsonify(message="client create/update {}".format(resp_text)),\
+        return jsonify(message="client create/update {}".format(resp_text)), \
             resp_code
 
     else:
--- a/rbd-target-api.py
+++ b/rbd-target-api.py
@@ -2777,8 +2777,8 @@ def pre_reqs_errors():
 
     if dist in valid_dists:
         if dist == 'rhel':
-            import platform
-            _, rel, _ = platform.linux_distribution(full_distribution_name=0)
+            import distro
+            rel = distro.version()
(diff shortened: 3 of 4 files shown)


dmick (commented):
You'll also need to add python3-distro to the spec "Requires:" section (it's not a builtin library, but it is distro-packaged).

lxbsz:
> You'll also need to add python3-distro to the spec "Requires:" section (it's not a builtin library, but it is distro-packaged).

Okay, just missed the spec. Fixed it.

dmick:
The package is called "python3-distro"

lxbsz:
> The package is called "python3-distro"

Please see https://packages.fedoraproject.org/pkgs/python-distro/.

Both the `python3-distro` and `python2-distro` are subpackages of the `python-distro`. I have tested it on both `centos` and `fedora` and it worked:

```
[xiubli@localhost src]$ sudo dnf reinstall python-distro
Last metadata expiration check: 3:52:47 ago on Thu 18 Apr 2024 05:50:21 AM CST.
Dependencies resolved.
==============================================================================================================================================================================================================================================================
 Package                                                          Architecture                                             Version                                                          Repository                                                   Size
==============================================================================================================================================================================================================================================================
Reinstalling:
 python3-distro                                                   noarch                                                   1.5.0-7.el9                                                      appstream                                                    37 k

Transaction Summary
==============================================================================================================================================================================================================================================================

Total download size: 37 k
Installed size: 153 k
Is this ok [y/N]: 
```

Could you forseee any other issue with this @dmick ?

dmick:
I don't understand why this centos 8.stream host is different:

[dmick@vossi03 tmp]$ dnf provides python-distro
Last metadata expiration check: 1:11:40 ago on Wed 17 Apr 2024 06:15:46 PM PDT.
Error: No Matches found
[dmick@vossi03 tmp]$ dnf search python2-distro
Last metadata expiration check: 1:11:57 ago on Wed 17 Apr 2024 06:15:46 PM PDT.
No matches found.
[dmick@vossi03 tmp]$ dnf search python3-distro
Last metadata expiration check: 1:12:05 ago on Wed 17 Apr 2024 06:15:46 PM PDT.
===================== Name Exactly Matched: python3-distro =====================
python3-distro.noarch : Linux Distribution - a Linux OS platform information API
========================= Name Matched: python3-distro =========================
python3-distroinfo.noarch : Parsing and querying distribution metadata stored in
                          : text/YAML files

lxbsz:
> I don't understand why this centos 8.stream host is different:
> 
> [dmick@vossi03 tmp]$ dnf provides python-distro Last metadata expiration check: 1:11:40 ago on Wed 17 Apr 2024 06:15:46 PM PDT. Error: No Matches found [dmick@vossi03 tmp]$ dnf search python2-distro Last metadata expiration check: 1:11:57 ago on Wed 17 Apr 2024 06:15:46 PM PDT. No matches found. [dmick@vossi03 tmp]$ dnf search python3-distro Last metadata expiration check: 1:12:05 ago on Wed 17 Apr 2024 06:15:46 PM PDT. ===================== Name Exactly Matched: python3-distro ===================== python3-distro.noarch : Linux Distribution - a Linux OS platform information API ========================= Name Matched: python3-distro ========================= python3-distroinfo.noarch : Parsing and querying distribution metadata stored in : text/YAML files

Okay, so my way won't work for all the cases. I will update it. Thanks for your feedback.

ceph-iscsi.spec
```diff
@@ -67,6 +68,7 @@ Requires:       python3-rtslib >= 2.1.fb68
 Requires:       python3-cryptography
 Requires:       python3-pyOpenSSL
 Requires:       python3-requests
+Requires:       python3-distro
```
idryomov, line 71: Nit: Put it in the same spot as the one for legacy Python (i.e. between `python3-setuptools` and `python3-rados` for consistency.

idryomov (commented):
LGTM, although I don't understand why we are also reading `/etc/os-release` manually instead of using `platform` (or, with this PR, `distro`) for all of it.  Apparently this has something to do with support for SUSE (https://github.com/ceph/ceph-iscsi/pull/107), but the issue that was being worked around isn't explained anywhere.

ceph-iscsi.spec
lxbsz: Done.

idryomov (approved):

PR #278, opened 2024-10-01
add Makefile for source RPM

ktdreyer:
Automate the commands to generate a source RPM. This makes it easier to build this package with "mock".


.gitignore
```diff
@@ -23,6 +23,7 @@ var/
 .installed.cfg
 *.egg
 non-git/
+*.tar.gz
```
idryomov, line 26: Should `*.rpm` be ignored is well?  I think `rpmbuild` would generate a `.src.rpm` file.

idryomov (approved):

PR #279, opened 2024-10-04
device_status: avoid AttributeError when logging debugging info

idryomov:
```
Exception in thread Thread-2:
  File "/usr/lib64/python3.6/threading.py", line 937, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/site-packages/ceph_iscsi_config/device_status.py", line 162, in run
    stats_dict = dev_status.get_status_dict()
AttributeError: 'NoneType' object has no attribute 'get_status_dict'
```

### DIFF
--- a/ceph_iscsi_config/device_status.py
+++ b/ceph_iscsi_config/device_status.py
@@ -157,6 +157,5 @@ class DeviceStatusWatcher(threading.Thread):
                         dev_status.changed_state = False
 
             # debugging info
-            dev_status = self.get_dev_status(image_name)
-            stats_dict = dev_status.get_status_dict()
-            self.logger.debug(stats_dict)
+            status_dict = {k: v.get_status_dict() for k, v in self.status_lookup.items()}
+            self.logger.debug("device status {}".format(status_dict))
--- a/ceph_iscsi_config/device_status.py
+++ b/ceph_iscsi_config/device_status.py
@@ -125,7 +125,6 @@ class DeviceStatusWatcher(threading.Thread):
             svc = json.loads(outb).get('tcmu-runner')
             if svc is None:
                 self.logger.warning("there is no tcmu-runner data available")
-                self.state = "Unknown"
                 continue
 
             image_names_dict = {}


lxbsz (approved):

lxbsz (approved):

ceph_iscsi_config/device_status.py
```diff
@@ -157,6 +157,5 @@ def run(self):
                         dev_status.changed_state = False
 
             # debugging info
-            dev_status = self.get_dev_status(image_name)
-            stats_dict = dev_status.get_status_dict()
-            self.logger.debug(stats_dict)
+            status_dict = {k: v.get_status_dict() for k, v in self.status_lookup.items()}
+            self.logger.debug("device status {}".format(status_dict))
```
lxbsz, line 161: @idryomov 

I checked the old code again, IMO we should move these debug code into the `for` loop instead ?

Currently it will always print the debug logs for  the image in the end of the `status_lookup` list, is this what we want it to be ?

idryomov: > I checked the old code again, IMO we should move these debug code into the `for` loop instead ?

Moving it into the loop would mean generating one log message per image -- we probably don't want that.

> 
> Currently it will always print the debug logs for the image in the end of the `status_lookup` list, is this what we want it to be ?

This is how it's been since 2020, except that it was broken and crashed thread in case the last image on the list had just been removed (and possibly also in case of no gateways/images?).  Honestly, I don't know -- as I said in the commit message the original intent isn't really clear.  Dumping the status for just one, usually the most recently added image doesn't make sense to me since it's unlikely to be all that useful in actual debugging.

lxbsz: I just checked the old PRs about this and I believe Mike wanted to debug all the existing images, not only the new one just added because other image states could be changed and we need to check them too.

And another case is that during the status showing there most possibly won't new images will be added, so in this case it will always show the debug info of the same image in the `status_lookup` list.

Mike just did this outside of the `for` loop by mistake.

idryomov: > I just checked the old PRs about this and I believe Mike wanted to debug all the existing images, not only the new one just added because other image states could be changed and we need to check them too.

I figured that that was probably the intent and preserved it in this PR (see how `status_dict` is generated).

> 
> And another case is that during the status showing there most possibly won't new images will be added, so in this case it will always show the debug info of the same image in the `status_lookup` list.
> 
> Mike just did this outside of the `for` loop by mistake.

I'm not sure I'm following -- are you suggesting to move the "debugging info" code into the for loop?  It seems like that would:
- take a bit more code because the loop has two branches which can't be handled uniformly since one of them removes the image from `status_lookup` dictionary
- generate one log message per image instead of one log message per `status_check_interval` tick for all images combined

lxbsz: I mean just move them to the `else:` branch in :

```
            for image_name in list(self.status_lookup):
                if image_names_dict.get(image_name) is None:
                    ...
                else:
                    ==> HERE <==
```

idryomov: I see.  But this
> - generate one log message per image instead of one log message per `status_check_interval` tick for all images combined

would still hold, do you think it's better?  `status_check_interval` defaults to 10 seconds and dumping X log messages every 10 seconds can be spammy.

lxbsz: IMO any will be okay. But to be safe I think we can just "generate one log message per image".

idryomov: > IMO any will be okay.

Merging as is based on the results of my test.

PR #282, opened 2026-03-26
iscsi: map Rocky Linux to RHEL in pre_reqs OS check

kginonredhat:
Rocky reports ID=rocky in /etc/os-release. pre_reqs_errors() only whitelisted rhel/suse/debian/ubuntu (plus centos→rhel), so /sysinfo/checkversions returned "OS is unsupported", valid_gateway failed, and gwcli gateway create left targets with no portals.

Fixes https://tracker.ceph.com/issues/75359
Signed-off-by: Kobi Ginon <kginon@redhat.com>


djgalloway:
ok to test

idryomov (approved):

